PR #31, opened 2017-09-20
Add logo image

ManfredKarrer:


cbeams:
This should probably go in `/press-material/logos/` where the rest of the logos are. Not really an ideal path, but that's what was already there in the site prior to migration to Jekyll. Looks like this new white-background image isn't referenced anywhere on the website (i.e. from any html files, etc). Can we just move it?

ManfredKarrer:
Its the link used for the logo on the GH readme file.

PR #42, opened 2018-02-08
Add security deposit information for BTC, LTC and DASH

ripcurlx:
I've added default, max & min values for security deposits (buyer, seller) for BTC, LTC and DASH. I left out DODGE as it is not selectable atm within the client.


cbeams (approved):
ACK, thanks! (And note, it's DOGE, not DODGE.)

ripcurlx:
Thanks for the DOGE hint 🐶 😅

PR #66, opened 2018-07-04
Update links, release notes and assets for v0.7.1 release

ripcurlx:
@cbeams please wait for my go before merging this PR!

PR #69, opened 2018-07-07
Extract 'version' variable on /downloads page

cbeams:
Previously, when releasing a new version, multiple error-prone changes
to the downloads page were required. Now, the `version` variable has
been extracted, meaning that it is necessary to update just one line in
the front matter of the page. See https://jekyllrb.com/docs/variables/
for details.

/cc @ripcurlx; see #66 to see what was required prior to this change.


cbeams:
I'm merging this straightaway, as I have additional changes to make. @ripcurlx, just have a look, so you understand how much easier changing this page will be next time we do a release.

PR #83, opened 2018-09-06
Set Android app as released

ripcurlx:
@joachimneumann just released the final version of the Android app


cbeams (approved):
ACK. (tested via deploy preview)

PR #85, opened 2018-09-10
Change app store button text to english

m52go:
To make text in the download button English instead of German.


ripcurlx:
ACK - good catch @m52go, I missed to update the svg part of the button.

cbeams (approved):
utACK. Good catch indeed. I should have caught this in my review yesterday as well. Thanks.

PR #90, opened 2018-10-20
Add redirect for /docs/dao/full-node

ManfredKarrer:


cbeams (approved):
ACK. Note, @ManfredKarrer, that I want to flatten out the docs page hierarchy, so that there are no pages like `/dao/full-node.html`, but rather just `/dao-full-node.html` or whatever. There are only a few pages that are nested within directories now, and they're mostly "legacy" docs. I think that immediately after we launch the redesign, we (me and/or @m52go) should go in and flatten everything out, update/add redirects accordingly. So this new redirect is fine in the meantime; just mentioning the above FYI.

ManfredKarrer:
Ah ok, I will update it with a flat link so we dont need to change it later...

PR #87, opened 2018-10-07
Remove outdated pages

ManfredKarrer:
- Remove crowdfunding, finance, governance pages as they are outdated.
- Remove contact and dao page as they did not had content just a link.
- Add new link in philosophy and remove one.
- Reorder community links by relevance.
- Remove link to keybase team

PR #91, opened 2018-10-20
Link to Phase Zero doc at docs.bisq.network

cbeams:
Instead of linking directly to the source on GitHub. This is a follow-up
to PR #87.


ManfredKarrer (approved):
utACK

PR #97, opened 2018-12-03
Add download links to os text

m52go:
1. Text such as "Download for Mac" and "Download for Windows" in download menus are now linked to downloads (per discussion with @ripcurlx).
2. Download links for 32-bit Windows are removed.
3. Removed trailing slash for root URL. It was causing double slashes in site links (e.g., bisq.network//roadmap).


ManfredKarrer (approved):
utACK

PR #99, opened 2018-12-04
Fix blank under 'get started' button on mobile

m52go:
Another small PR to fix 2 small issues:
1. On mobile, there's currently a blank under the Get Started button. This PR adds fallback text.
2. It removes the 32-bit Windows download link from the actual /downloads page (forgot this in my last PR).


ManfredKarrer (approved):
utACK

PR #102, opened 2018-12-05
Rework download links

m52go:
I found issues with download links:

- primary download links did work in firefox at all (!)
- ios showed ios and macos text at the same time
- inconsistent os highlighting in download menus (windows and macos particularly)
- add fallback text for non-windows-mac-linux users

Other notable changes:

- re-add analytics (the ga initialization script was missing before)
- remove all 32-bit links and rename 64-bit links (the '64 Bit' text from 3d23ce25b0ff8e080987076ddf467c0436720764 is gone with this PR so that should be an easy merge conflict)

So I spent some time yesterday reworking the code and UI for download links. While seemingly minor I figured it worth the time because getting users to download the Bisq software is probably our #1 goal for the website, and it should be a flawless experience.

@ripcurlx if you or someone with access to Analytics can verify that analytics are working again, that would be great. I was getting ga errors before, and I'm not getting them any more, but I can't be 100% sure events are actually being recorded on the server.

### DIFF
--- a/_includes/homepage_content.html
+++ b/_includes/homepage_content.html
@@ -26,7 +26,7 @@
                                     Download for Windows
                             </div>
                             <div class="float-right">
-                              <a href="https://github.com/bisq-network/bisq/releases/download/v{{ site.client_version }}/Bisq-64bit-{{ site.client_version }}.exe">64bits</a>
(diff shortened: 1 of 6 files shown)


ripcurlx:
@m52go Looks good to me in the preview! But it seems that you have a conflict that needs to be resolved.

ripcurlx:
ACK

PR #103, opened 2018-12-10
Fix markets page

m52go:
There were several errors in the markets graphs and tables on the Markets section with respect to units, prices, and aggregation (averages showing instead of sums).

I also fixed some aesthetic items, such as limiting significant figures (e.g., so decimal numbers don't have 15 digits).

The charting plugin is rather sophisticated so I didn't mess with it too much, but most other code was heavily revised/refactored.

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -757,6 +757,18 @@ ol li {
 
 /****************/
 
+
+/*****************
+highcharts chart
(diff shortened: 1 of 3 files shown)


ManfredKarrer (approved):
utACK

PR #104, opened 2018-12-10
Update features copy

m52go:
This updated Features text on the front page better reflects aspects of Bisq's decentralization, security,
and openness.

I will add a link to DAO materials under this section as soon as our first user-facing DAO doc is out (which will be soon, as the first voting cycle just ended).

Minor: I also changed the survey link in _redirects to https.


ManfredKarrer (approved):
utACK

PR #108, opened 2019-01-02
Link to DAO docs from front page

m52go:
This PR adds links to user-facing DAO docs on the front page. 

It also includes a short blog post about DAO going live on testnet, and because it links to docs from [this PR](https://github.com/bisq-network/bisq-docs/pull/96), **this PR should only be merged after that PR is merged.**


ManfredKarrer (commented):
utACK

PR #117, opened 2019-01-30
[WIP] Add dao page to website

m52go:
This proposed web page is the new tool I suggest should be the first place we send people to learn about the DAO. It hits the most common objections people have about new tokens & the Bisq token specifically. It includes new visuals, video, an interactive simulation, FAQs, and an organized list of links to all the resources we've created. 

It's visual and interactive and simple (but thorough), targeting a variety of learning styles, so people can learn however they learn best.

The simulation is particularly effective (in my opinion) because it can also be made into a GIF for easy sharing, used on stage while speaking, etc. It's the best way I can think of to explain the primary need for the DAO in under 30 seconds.

I'm making the PR now for feedback on everything but the video (existing one is just a placeholder for ithe next ~1 day until I get a new one done). Feedback welcome.

**To clarify, existing video is a placeholder**. Everything else is done/ready for review.


m52go:
You can preview the site here:
http://stevejain.com/playground/bisq-dao/dao/

Netlify seems to assume wrong site-url so its preview is not correct. 

All I did to get the site to work on my domain is set `url: "http://stevejain.com/playground/bisq-dao"` in `_config.yml`. Of course you can build from source to be 100% sure.

m52go:
My approach was off with this one. I think there's a lot of great material to salvage from it, but the final product will be very different from this PR, different enough to warrant a fresh PR.

PR #120, opened 2019-02-18
Fix missing css animation code

ripcurlx:


m52go (approved):
Looks good to me. I don't think these vendor-specific prefixes are usually needed anymore but better safe than sorry.

PR #122, opened 2019-02-20
Add URL redirects to GitHub content

devinbileck:
URL redirects for downloads, compensation requests, and
proposals; all of which currently go to GitHub.

See https://github.com/bisq-network/bisq/issues/2429


ManfredKarrer (approved):
utACK

PR #128, opened 2019-02-25
Load JS async to speed up page loading times

ripcurlx:
I checked with our GA data on browser that we need to support and I found only one case where it is supported partial (IE 8)
https://caniuse.com/#feat=script-defer


m52go (approved):
ACK

PR #137, opened 2019-03-04
Fix blog post excerpt with disclaimer

m52go:
Blog post disclaimers are currently showing up in the excerpts on the blog listing page.

This PR explicitly sets an `excerpt` property in 'front matter' attributes at the top of the post to avoid this.


ManfredKarrer (approved):
utACK

Issue #139, opened 2019-03-07
Long sat values on trades (ie: .00000004) show up abbreviated. Is this by design?


nihirisuto:
See screenshot. I think this is confusing personally. 
<img width="1136" alt="screen shot 2019-03-06 at 6 43 06 pm" src="https://user-images.githubusercontent.com/32186637/53928754-1a9b9800-4040-11e9-8613-d6ceb025d114.png">

HarryMacfinned:
Are you sure it was .00000004 = 4x1e-8 at this time ?

ghost:
You are right, it should be 4x1e-8. 
Or 5x1e-8 right now.

m52go:
@chris-larsen thanks for pointing this out. It's not by design. Fix incoming.

m52go:
Merged, closing. Let us know if you still see issues.

PR #141, opened 2019-03-07
Make markets decimal handling more consistent

m52go:
Looks like JavaScript displays numbers with 6 digits or fewer after the decimal differently from those with more than 6 digits after the decimal.

This fix makes all small numbers appear more consistently, so that exponential notation does not show.

Addresses #139.


ManfredKarrer (approved):
utACK

PR #144, opened 2019-03-08
Add headers to improve security

ripcurlx:
Based on feedback from https://observatory.mozilla.org/analyze/bisq.network.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -12,6 +12,7 @@ feed:
   path: blog/feed.atom
 include:
   - _redirects
+  - _headers
 exclude:
   - CNAME
   - README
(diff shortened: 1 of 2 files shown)


ripcurlx:
It is configured as described in https://www.netlify.com/docs/headers-and-basic-auth/. But I don't see it in the test build.

ripcurlx:
I do see the correct headers on sub sites (e.g. vision), but it should be actually also configured for the root as well.

ripcurlx:
To use https://infosec.mozilla.org/guidelines/web_security#content-security-policy is a little bit more effort as I would need to check for all in-line scripts and do lots of testing not to break anything. Will be a next step.

m52go:
ACK

PR #145, opened 2019-03-08
[WIP] Add headers security configuration for root domain on netlify

ripcurlx:


ripcurlx:
It does work - I missed that the root domain returned with http 304 in some cases having still the old http headers.

PR #150, opened 2019-03-18
Change redirect for bsq genesis instructions

m52go:


m52go:
To go with https://github.com/bisq-network/bisq-docs/pull/120.

Issue #147, opened 2019-03-14
Create stats page on website


m52go:
To give people an idea of network activity without having to wade into the spreadsheet.

Proposed metrics:
- overall volume
- fees
- bsq price + supply
- commits (?)
- number of contributors (?)

Anyone with ideas on which metrics should or shouldn't be included, please add.

huey735:
I think that it'd be useful to also have a table/spreadsheet of the cost of a trade on Bisq:
* the trading fee for makers and takers
* the security deposit for makers (not necessary since it's not set in stone but rather set by the offer 
maker)
* nº of on chain transactions needed to finalize the trade for which the maker and taker are responsible for

ManfredKarrer:
Trade fee for maker is 0.1% and for taker 0.3%.
Sellers security deposit is 0.5% of trade amount and buyers deposit default 2% but can be changed by maker.
The maker pays 1 times the miner fee for the maker fee tx. For the other 3 txs (taker fee tx, deposit tx and payout tx) the taker pays the miner fee.

I think that should be in the FAQ (I think it is but not sure if updated) but not subject to the stats page.

huey735:
> Sellers security deposit is 0.5% of trade amount and buyers deposit default 2% but can be changed by maker.

@ManfredKarrer I think you meant offer-makers and offer-takers instead of sellers and buyers. We probably should promote this vocabulary instead. I think it's less confusing to understand from one's point of view if one's taking an existing offer or making a new one. On the other side it's easy to confuse whether one's buying or selling, because in any trade we're both buying and selling something. Also is the security deposit for offer-makers 0.5% of trade amount or 0.005btc no matter the trade amount?

It'd be useful to have those trades info plus trade limits in a table that'd be updated as changes are made:

**Fees per trade**

Categories | Offer-maker | Offer-taker | Version
------------ | ---------------- | ------------- | -------------
Trading fee | 0.1% of trade amount | 0.3% of trade amount | v.0.95
Security deposit* | 0.005btc | decided by offer-maker | v.0.95
Nº of miner's fee** | 1 | 3 | v.0.95


*this value is just a collateral and it's reimbursed to the traders and the end of trade.
**these numbers don't include the deposit and withdraw from and to an external wallet. 

**Trade Limits**

Payment method | 1st month (25%) | After |  Version
------------ | ---------------- | ------------- | -------------
Altcoins | 0.25btc | 1btc | v.0.95
[...]

PR #156, opened 2019-03-29
Add stats page

m52go:
To address https://github.com/bisq-network/bisq-website/issues/147.

Notes:

* Added drop-down to navbar to make it less crowded
* Data is hard-coded for the moment, to be updated monthly (with first update right after this upcoming voting cycle is done)
* Bottom blocks need to be centered

Also looks like there are some merge conflicts at the moment (will fix).

But please review for substance.

Preview is broken, so here are screenshots.

Screenshot from 2019-03-29 11-38-43
Screenshot from 2019-03-29 11-39-53


ManfredKarrer:
Date of downloads is wrong: Released Aug 22nd, 2018
Also downloads number contains all sig files. Would be good to get those filtered out. A small webservice delivering the pure app downloads mighty be nice.

ManfredKarrer:
Not sure if a more graphical approach would be better? @pedromvpg would you have time to make a suggestion?

m52go:
@ManfredKarrer I used 0.9.3 because it had the best numbers...took numbers from here and added up only .dmg .exe .deb numbers:

Screenshot from 2019-03-29 11-42-11

We could use a more recent version but numbers are much lower.

Yes automating these figures will be great, was thinking we could do it all at once, when we automate the Bisq stats.

pedromvpg:
Stats page
https://www.figma.com/proto/D9XHYXcNcobektGe5DpzAhVl/New-Pages?node-id=3%3A1&viewport=-664%2C-169%2C0.357336&scaling=min-zoom

Is this the right list of data to be displayed? If not is there a consolidated list I can reference?

Stats-Desktop

m52go:
That looks good to me. Generally I think we want to include as much as we can from the [analytics infographic](https://bisq.slack.com/archives/C7WRK8DBM/p1551964649007000) and whatever else makes sense (e.g., link to the network monitor).

ManfredKarrer:
@pedromvpg figma crashes often my firefox tabs.... the adobe page was working better

ripcurlx:
I like the design 👍 . Regarding KPIs, graphs and naming I would suggest following:
- Let's name it `Statistics` instead of `Network stats`? It reminds me atm of real network monitoring and not Bisq trading statistics.
- [Graph] Volume
- [KPIs Trading/Client] Total Trades - Daily Average Trades (last month) - Downloads latest (major) release
- [Graph] BSQ Supply
- [Graph] BSQ Burned (by trading)
- [KPIs Community] Github Contributors - Twitter Followers - Slack Users

I think for a first version this information should be sufficient to spark interest.

m52go:
I like @ripcurlx's suggestions...very well thought-out.

Only thing I would mention is that it makes more sense to me to keep BSQ supply and created/burned on the same graph. Maybe the line indicates the net supply over time, and double-bars at each time interval indicate issuance and burning (not sure of the exact graph details; we can figure them out when we actually have the data to make the complete graph).

m52go:
I applied Pedro's design and Christoph's KPIs suggestions. I kept the order of items the same as Pedro's design, as alternating between graphs and KPIs looks a little strange, I think, especially on smaller screens where everything becomes a single column.

Also -- I stuck with v0.9.5 for downloads because v0.9 didn't have too many (which is what I assume ripcurlx meant by major release - 0.8, 0.9, 1.0, etc). I think we just need to play this by ear and use the version that makes most sense.

If there are no further suggestions, this should be ready to go.

Screenshot from 2019-04-12 08-14-17

PR #164, opened 2019-04-12
Update redirects

devinbileck:


m52go:
Is there a chance we could add some kind of identifier in these URLs to act as a sort of 'namespace' (so to speak) that refers to their purpose?

Like `https://bisq.network/dao-compensation/` instead of `https://bisq.network/compensation/`.

Only because it would be nice to be able to link to use clean URLs like `https://bisq.network/compensation/` for user-facing documentation/materials.

Not super critical, but thought I'd suggest it.

devinbileck:
Good point. Just not sure which is best?
`/dao-compensation`
`/compensation-requests`
...

And perhaps these need to be changed as well?
`/proposals`
`/roles`
`/reimbursement`

devinbileck:
@ManfredKarrer I guess it is going to cause problems if I change the links to `/dao-roles` in BondedRoleType. I did it on my local and my DAO state seems to be out of consesus now. So I will have to leave `/roles` as is.

ManfredKarrer:
Hm, not clear to me why that could cause a consensus issue, but don't have the time atm for lookinng into that. 
Regarding proposal: We use that atm for kind of BIP style proposals which are different to DAO proposals. Would like to avoid to mix it not inflate the BIP style proposals. Maybe we should rename the proposals to something different (BisqIP)?

devinbileck:
There are different types of proposals? Don't the BIP-style proposals now require voting upon via the DAO?

devinbileck:
Something like BEP Bisq Enhancement Proposal and BDP Bisq DAO Proposal?

devinbileck:
I attempted to change the links to `/dao-roles` in BondedRoleType again but this time did not encounter consensus issues with the DAO state. So appears to be ok to change those.

ManfredKarrer (approved):
utACK

ManfredKarrer:
> I attempted to change the links to /dao-roles in BondedRoleType 

I prefer to not change that. Not worth the risk.

Issue #166, opened 2019-04-16
Markets page price wrong


ManfredKarrer:
The displayed altcoin price seems to be inverted.
<img width="2064" alt="Screen Shot 2019-04-16 at 17 43 57" src="https://user-images.githubusercontent.com/1449498/56248879-64c85e00-606f-11e9-8726-9eb384125e26.png">

ManfredKarrer:
As I assume https://coinmarketcap.com/exchanges/bisq/ uses the same API it seems an issue on the webpage side not on the markets server.

m52go:
Yes this has been an issue for a while, but when I investigated earlier, fix was not trivial so I put it off. Will look at this week.

ManfredKarrer:
I think price is inverted atm. 1 / price is probably correct. If still not then it might be precision. Fiat has precision 4 and altcoins 8.

m52go:
Closing as complete, as this issue was fixed with #171.

PR #171, opened 2019-04-19
Fix markets tables and charts

m52go:
All pricing for altcoins is switched to BTC. 

I cross-checked all values with actual data, so everything should be right now, but please review carefully so we can make sure it's right this time.

Addresses https://github.com/bisq-network/bisq-website/issues/166.

Best would be to test locally, but here are screenshots:

Screenshot from 2019-04-19 12-05-20
Screenshot from 2019-04-19 12-04-49
Screenshot from 2019-04-19 12-04-58


ManfredKarrer:
@pedromvpg  @ripcurlx Could you review that that, not my domain...

ripcurlx (approved):
ACK

PR #187, opened 2019-06-04
Make minor edits for translations

m52go:
Most of these updates reflect changes discussed with @arunasurya on [this Google Doc](https://docs.google.com/spreadsheets/d/1P4JMLrcRtSWkxfh9jG7AXkfdgdkEYwgttGgly-ercXc/edit#gid=1623394417).

I did edit the Vision page a bit, as it seemed to have been written before Bisq was launched.


arunasurya (commented):
Looks great! The only mistake I see is repeated "the" in line 15, as in "the the primary way to...".

Issue #125, opened 2019-02-21
Use Red Hat or Fedora inspired image for the rpm download item in the dropdown


devinbileck:
Would be nice to have a Red Hat or Fedora inspired image for the rpm download item in the dropdown, rather than using the generic Linux icon:
![image](https://user-images.githubusercontent.com/603793/53206630-58330680-35e6-11e9-99eb-d46b405d0f99.png)

huey735:
@devinbileck what do you think of this?

![redhat-g](https://user-images.githubusercontent.com/44140017/59134448-21a1b180-8973-11e9-88a8-fed4916d19e3.jpg)

devinbileck:
Not bad. But perhaps should take inspiration from their new simplified icon with just the hat?
![image](https://user-images.githubusercontent.com/603793/59138955-ea1c1080-8944-11e9-96ca-4b661f540b70.png)

devinbileck:
Also, perhaps we should redo the Ubuntu icon as a Debian icon so as to represent the base distro from which Ubuntu is based on (or other distros that are based on it). For example, taking inspiration from:
![image](https://user-images.githubusercontent.com/603793/59139002-323b3300-8945-11e9-9265-8b4bcfa79a42.png)

And while we are at it, may want to do an Arch icon based on:
![image](https://user-images.githubusercontent.com/603793/59139005-38c9aa80-8945-11e9-9bfc-5071464404b2.png)

huey735:
Take a look. I prefer Ubuntu's logo over Debian's as the positive-negative space are more balanced.
![linux-distros-2](https://user-images.githubusercontent.com/44140017/59141519-25026080-89a6-11e9-8ac7-8151df439bfd.png)

devinbileck:
Looks good!
I do agree regarding the positive-negative space. However, I like the idea of using Debian for consistency as the base distro. Although, I suppose we could use Ubuntu and switch to the Fedora logo for the most "popular" distros that use those packages?

huey735:
I blended the dark and light blue of the original logo to limit the color scheme to only two colors.
![fedora](https://user-images.githubusercontent.com/44140017/59275399-26f14b80-8c54-11e9-8a2a-5d2fbfa51c8a.png)

devinbileck:
Looks good to me!

m52go:
Is the Fedora logo a different shade of green than the icons above it? It looks lighter.

[...]

PR #189, opened 2019-06-11
Update Linux distros icons

huey735:
Changing icons from OS selector drop-down menu to be more specific, as discussed [here](https://github.com/bisq-network/bisq-website/issues/125)
I also updated the code on _includes/homepage_content.html although I don't fully understand its purpose.

os-selector-dropdown-menu


m52go:
ACK

PR #192, opened 2019-07-05
Update stats page for june

m52go:
Will merge later today if there are no concerns. This is a straightforward numbers update.


huey735 (approved):

devinbileck:
Forgot to update the dates
image

m52go:
Ah damn. Looks like that's been neglected for a while. I think that label was useful before the DAO launched, but now it's probably not very useful. I will submit a PR to replace it with something better.

PR #193, opened 2019-07-05
Change stats page graph subtitles

m52go:
To address @devinbileck's comment [here](https://github.com/bisq-network/bisq-website/pull/192#issuecomment-508804162). 

Having the current month in the subtitle isn't a great idea as it's not an obvious thing for a maintainer to update...and for users, it's not a very meaningful subtitle. 

I think these are better.

### DIFF
--- a/js/stats.js
+++ b/js/stats.js
@@ -26,7 +26,7 @@ $( document ).ready( function() {
               y: 15
           },
           subtitle: {
-              text: 'As of March 2019',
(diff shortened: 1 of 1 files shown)


huey735:
I'd go further and say those aren't even necessary.

erciccione (approved):
Definitley better IMHO :)

devinbileck (approved):
utACK

PR #197, opened 2019-07-22
Fix language selector

m52go:
This pull request fixes the language selector to be (1) more responsive on desktop and (2) available on mobile.

It also fixes a minor bug with the scrolling animation on the DAO page.


m52go:
Going to merge this now as it's just HTML and CSS markup, and it's holding up more languages from being added. I've thoroughly tested this across browsers and screen widths and it seems to works fine.

PR #199, opened 2019-07-29
Updates for v1.1.4 release

ripcurlx:


huey735 (changes requested):
There's now a /pt-PT/downloads.html file that also needs to have it front matter updated. It's unfortunate that Jekyll doesn't allow us to simply reference the site variable 'client_version' in the front matter.

ripcurlx:
> There's now a /pt-PT/downloads.html file that also needs to have it front matter updated. It's unfortunate that Jekyll doesn't allow us to simply reference the site variable 'client_version' in the front matter.

Good to know - I'll update the file as well.

PR #200, opened 2019-08-06
Update stats page for july 2019

m52go:


m52go:
Merging, as this is a simple figures update.

PR #201, opened 2019-08-06
Try pt-pt image fix

m52go:
Hard to test locally, as images work fine locally, and I don't have access to Netlify. Will revert this pull request if it doesn't work.


m52go:
It worked! @huey735

huey735:
Just to be sure, does this mean that for the future we need to rename the image files and the urls inside dao_content_tr.yml?

m52go:
Yes, each image for each language needs to have a filename that's different from that of the same image in every other language (e.g., just append the image file with the language code).

To be honest, I'm not sure it's worth the effort. Feel free to change them all back English if you like. I don't think it's a big deal. The images are meant to be visual anyway, so I'm sure people could get the main idea from the illustrations.

But that's the way to make it functional, if you'd like to stick with it.

PR #204, opened 2019-08-08
Update stats and download version

huey735:
@ripcurlx there are now 4 additional download pages on the website:
* /de/downloads.html
* /es/downloads.html
* /pt-PT/downloads.html
* /zh-CN/downloads.html

@m52go going forward, please also update the /_data/strings_en.yml when doing changes like updating the stats. The strings inside /_data/strings_en.yml are distributed into the following lists:
* footer
* homepage_content
* main_nav
* os_selector_options:
* dao
* downloads
* faq
* markets
* stats

I'll be setting up Transifex so that the "strings" resource updates when there are changes to  /_data/strings_en.yml.

### DIFF
--- a/_includes/main_nav_tr.html
+++ b/_includes/main_nav_tr.html
@@ -1,7 +1,14 @@
  <nav class="navbar navbar-expand-lg navbar-light bg-light fixed-top pl-md-5 pr-md-5 pl-sm-3 pr-0">
 
-    <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+    <div>
+      <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
 
+      <!-- Outdated translation tag-->
+      {% if page.lang == page.lang == "de" or page.lang="es" or "pt-PT" or page.lang == "zh-CN" %}
+        <span class="navbar-text" style="color: orange;">Outdated translation</span>
+      {% endif %}
+    </div>
+      
     <!-- Language selector for mobile -->
     {% if page.lang %}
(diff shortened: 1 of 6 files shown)


_includes/main_nav_tr.html
```diff
@@ -1,7 +1,14 @@
  <nav class="navbar navbar-expand-lg navbar-light bg-light fixed-top pl-md-5 pr-md-5 pl-sm-3 pr-0">
 
-    <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+    <div>
+      <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
 
+      <!-- Outdated translation tag-->
+      {% if page.lang == page.lang == "de" or page.lang="es" or "pt-PT" or page.lang == "zh-CN" %}
```
m52go: When is this "outdated translation" text supposed to show? This line seems to indicates it will show any time the current language is not EN (and indeed that's what happens when I run this locally).

Otherwise, please double-check the logic. `page.lang == page.lang` is redundant and you're missing the equality operator for ES (`page.lang="es"` needs `==`).

m52go (changes requested):
Agreed about the stats numbers—will remember this next time.

That `version` front-matter variable is not used...the `site.client_version` is the way this page gets the version number. I'll make a PR to remove it altogether (done; see #206). Then we won't have to worry about updating those files at all :)

_includes/main_nav_tr.html
huey735: My idea is that we'd regularly update it by taking out and leaving the relevant ones.

huey735:
Closed because https://github.com/bisq-network/bisq-website/pull/217 and https://github.com/bisq-network/bisq-website/pull/218 are a better PRs.

PR #205, opened 2019-08-11
Fix mobile notifications links on home page

m52go:
The iOS link on the home page was going to the Google Play store.

@huey735 I also shortened some of the texts for the mobile app download buttons. Could you check them to make sure they look ok? For Portugese for example, the text may not be gramatically perfect, but do you think it will work?

**Before**

`iosNotificationApp: Baixar App de Notificação para iOS`

Screenshot from 2019-08-11 16-14-03

**After**

`iosNotificationApp: iOS App de Notificação`

Screenshot from 2019-08-11 16-13-48

### DIFF
--- a/_data/de.yml
+++ b/_data/de.yml
@@ -19,8 +19,8 @@ homepage_content:
   buttonGetBisq: Bisq erhalten
   buttonToggle: Dropdown-Menü umschalten
   downloads: Downloads
-  androidNotificationsApp: Android Benachrichtigungs-App
-  iosNotificationApp: iOS Benachrichtigungs-App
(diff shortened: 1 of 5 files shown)


huey735 (commented):
In Portuguese "App de Notificação IOS" and "App de Notificação Android" is better. I assume it's similar in Spanish but can't be sure. Please also update the strings inside _data/strins_en.yml.

m52go:
@huey735 thanks, updated. The strings in `_data/strings_en.yml` don't change. This change was merely to shorten the existing strings so they fit on mobile screens (original meaning stayed the same).

huey735 (approved):
utACK for es and pt-PT

PR #206, opened 2019-08-21
Remove extraneous `version` variable on downloads page

m52go:
Variable wasn't being used.

FYI @ripcurlx...there's no need to update `download.html` any more for new releases. That page gets the software version from `_config.yml`.


huey735 (approved):
utACK

PR #208, opened 2019-08-22
Replace Google Analytics with self-hosted Matomo (fixes #191)

### LINKED ISSUE #191: [Proposal] Get rid of google analytics
I think Google Analytics could be replaced with another more privacy-friendly service. For getmonero.org we use [Matomo](https://matomo.org/), which is self-hosted and open source. I think the bisq users would appreciate the switch :)

wiz:


m52go:
Looks good to me, as long as events are hitting the server side alright. I'm running this locally and not seeing any errors in the console. 

@wiz ready to merge?

wiz:
Yeah go for it :+1:

PR #210, opened 2019-08-23
Enable HSTS preload for bisq.network domain name

wiz:
HSTS is already enabled, but if we add "preload" we can add Bisq to https://hstspreload.org/


m52go:
utACK

PR #212, opened 2019-08-23
Fix site headers on blog post pages

m52go:
Right now, site headers do not render properly on blog post pages. 

See example:
https://bisq.network/blog/bisq-dao-for-bitcoin-maximalists/

Fix is to add a `ref` property in the front matter of posts so that they are rendered with the correct site header (`homepage_content.html` instead of `homepage_content_tr.html`).


huey735 (approved):
ACK

I've ran this locally and it works. But can you help me understand how is Jekyll serving the page?

Is it through this lines in _layouts/default.html
```
<!--Select includes/main_nav according to page lang or absence of lang-->
          {% if page.lang == "en" or page.ref == "blog" %}
            {% include main_nav.html %}
          {% else %}
            {% include main_nav_tr.html %}
          {% endif %}


            {% if page.url == '/' %}

                <div class="outter">
                    {% include homepage_content.html %}
                </div>

            {% else %}

                <div class="outter">
<!-- Select includes/homepage_content according to page lang -->
                  {% if page.ref == "index" %}
                    {% include homepage_content_tr.html %}
                  {% else %}
                    {{ content }}
                  {% endif %}
                </div>

            {% endif %}

<!-- Select includes/footer according to page lang -->
              {% if page.lang == "en" or page.ref == "blog" %}
                {% include footer.html %}
              {% else %}
                {% include footer_tr.html %}
              {% endif %}
```

m52go:
> Is it through this lines in _layouts/default.html

Yes, exactly. Before adding the `ref: blog` front matter to posts, post pages met neither of [these conditions](https://github.com/bisq-network/bisq-website/blob/master/_layouts/default.html#L736) to show ` main_nav.html`, so ` main_nav_tr.html` was rendered instead, which caused the problem.

PR #213, opened 2019-08-23
Add French translation

huey735:
Building on top of https://github.com/bisq-network/bisq-website/pull/204

### DIFF
--- a/_data/strings_en.yml
+++ b/_data/strings_en.yml
@@ -303,26 +303,26 @@ stats:
   pBSQDistribution: Mainnet BSQ distribution is slated for April 2019.
   pTotalTrades: >
     <p class='stat-label'>Total Trades</p>
-    <p class='stat-figure'>32,318</p>
+    <p class='stat-figure'>34,535</p>
     <p class='stat-note'>Through June 2019</p>
   pDailyAverage: >
     <p class='stat-label'>Daily Average Trades</p>
-    <p class='stat-figure'>94</p>
-    <p class='stat-note'>For June 2019</p>
+    <p class='stat-figure'>72</p>
+    <p class='stat-note'>For July 2019</p>
   pSoftwareDownloads: >
     <p class='stat-label'>Software Downloads</p>
-    <p class='stat-figure'>13,723</p>
+    <p class='stat-figure'>17,968</p>
     <p class='stat-note'>For v1.1.2 (approximate)</p>
   pGithubContributors: >
     <p class='stat-label'>GitHub Contributors</p>
-    <p class='stat-figure'>172</p>
-    <p class='stat-note'>As of June 2019</p>
+    <p class='stat-figure'>182</p>
+    <p class='stat-note'>As of July 2019</p>
   pTwitterFollowers: >
     <p class='stat-label'>Twitter Followers</p>
-    <p class='stat-figure'>17,588</p>
-    <p class='stat-note'>As of June 2019</p>
+    <p class='stat-figure'>17,945</p>
+    <p class='stat-note'>As of July 2019</p>
   pSlackUsers: >
     <p class='stat-label'>Slack Users</p>
-    <p class='stat-figure'>1,183</p>
-    <p class='stat-note'>As of June 2019</p>
+    <p class='stat-figure'>1,258</p>
+    <p class='stat-note'>As of July 2019</p>
   pNetworkLoad: "<p class='monitor-note'>For P2P network load, Tor metrics, and other network data, please visit the <a href='https://monitor.bisq.network' target='_blank'>Bisq Network Monitor</a>.</p>"
--- a/_data/strings_en.yml
+++ b/_data/strings_en.yml
@@ -304,7 +304,7 @@ stats:
   pTotalTrades: >
     <p class='stat-label'>Total Trades</p>
     <p class='stat-figure'>34,535</p>
-    <p class='stat-note'>Through June 2019</p>
+    <p class='stat-note'>Through July 2019</p>
   pDailyAverage: >
     <p class='stat-label'>Daily Average Trades</p>
     <p class='stat-figure'>72</p>
--- a/_includes/main_nav_tr.html
+++ b/_includes/main_nav_tr.html
@@ -1,7 +1,14 @@
  <nav class="navbar navbar-expand-lg navbar-light bg-light fixed-top pl-md-5 pr-md-5 pl-sm-3 pr-0">
 
-    <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+    <div>
+      <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
 
+      <!-- Outdated translation tag-->
+      {% if page.lang == page.lang == "de" or page.lang="es" or "pt-PT" or page.lang == "zh-CN" %}
+        <span class="navbar-text" style="color: orange;">Outdated translation</span>
+      {% endif %}
+    </div>
+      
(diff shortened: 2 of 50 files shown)


m52go:
@huey735 the French translation looks good. But I can't merge this PR for the issue noted below.

The outdated translations alert tool still doesn't seem right to me. In my testing, it appears on any non-English page no matter what. If I understand your intentions correctly, what you probably want instead are flags for each language (implemented as page variables) that can be toggled every time an update is made to the corresponding English page. Then the outdated-translation text would show on every non-English page until it's updated with the translation, and the flag variable can then be turned off again.

If you want, I can implement this for you.

FYI in the future, it might be best to make separate PRs for separate objectives so at least the good commits can be merged.

huey735:
@m52go it looks like this right now
```
      <!-- Outdated translation tag-->
      {% if page.lang == "de" or page.lang= "es" or "pt-PT" or page.lang == "zh-CN" %}
        <span class="navbar-text" style="color: orange;">Outdated translation</span>
      {% endif %}
    </div>
```
My idea is to add or remove "page.lang == langcode" depending on what is what's not up to date. Maybe there's a better way to do it.
I think you want to make it more specific to a page? Like if only the vision version of the Portuguese page is outdated, to only show it only on that page?
Maybe it's best to go ahead and remove it for now.

I'm still learning Github. I thought that in order to do two concurrent but separate PRs I needed to work on two different branches but that doesn't seem like the case. I guess I need to work on different forks. I'll improve next time.

m52go:
> My idea is to add or remove "page.lang == langcode" depending on what is what's not up to date.

Ah, I see now. Generally it's bad practice to edit code to achieve that kind of functionality. Rather, it's better to structure the code so that changes can be made in the front end (in this case, through Liquid page variables). 

Another benefit of this approach: by using page variables, you could mark specific pages as outdated. Otherwise, marking a language as outdated makes it outdated through the whole site, even if it's just one page that needs updates.

> I thought that in order to do two concurrent but separate PRs I needed to work on two different branches but that doesn't seem like the case. I guess I need to work on different forks.

No worries. You can make as many branches as you want :)

m52go:
So, concretely, I would do something like this.

Make an `outdated_translation` tag in the front matter of each page. Set it to `false`.

Then make the condition look like this:

```
{% if page.outdated_translation %}
    <span class="navbar-text" style="color: orange;">Outdated translation</span>
{% endif %}
```

Then whenever an update is made to an English page, all the non-English pages should have their `outdated` tags set to `true`, so the user sees the outdated message. Then, as the translations are updated, the flags are set back to `false`, and the message goes away.

huey735:
>No worries. You can make as many branches as you want :)

Yeah but this 'addfrench' branch of mine also inherited the commits of the 'updatetranslation' branch from https://github.com/bisq-network/bisq-website/pull/204. I need to figure out how to keep them separate.

m52go:
Oh I see. You could `git checkout master` and then make a new branch from there.

huey735:
Updated.
I guess I can close https://github.com/bisq-network/bisq-website/pull/204, right?

_includes/main_nav_tr.html
```diff
@@ -4,11 +4,11 @@
       <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
 
       <!-- Outdated translation tag-->
-      {% if page.lang == page.lang == "de" or page.lang="es" or "pt-PT" or page.lang == "zh-CN" %}
+      {% if page.lang == "de" or page.lang= "es" or "pt-PT" or page.lang == "zh-CN" %}
```
wiz: did you test this?

m52go:
@huey735 looks like some pages don't have the variable? For example pt-pt/index and pt-pt/vision? I didn't check thoroughly, but those stood out. There may be others.

If it isn't too much work, could you make the variable `outdated_translation` instead? Will be clearer for future contributors to understand what it means. Sorry. I updated my comment above but I guess it was too late.

_includes/main_nav_tr.html
huey735: That should have been updated with the latest commit as it's different now.
https://github.com/huey735/bisq-website/blob/addfrench/_includes/main_nav_tr.html

huey735 (commented):
@wiz _includes/main_nav_tr.html has been properly updated. For some reason it didn't show in the previous commit.

_includes/main_nav_tr.html
```diff
@@ -1,6 +1,13 @@
  <nav class="navbar navbar-expand-lg navbar-light bg-light fixed-top pl-md-5 pr-md-5 pl-sm-3 pr-0">
 
-    <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+    <div>
+      <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+
+      <!-- Outdated translation alert tool tag-->
+      {% if page.outdated %}
```
m52go: wiz did have a point though...you should really test. This variable also needs to be updated...would be an easy catch with a simple test, as it wouldn't work as intended.

huey735: Sorry guys. I'll get to it.

huey735 (commented):
Let's hope this is it. Thanks @wiz and @m52go for checking me.

wiz:
Can you deploy this to a netlify instance so we can see a preview?

_data/strings_en.yml
```diff
@@ -303,26 +303,26 @@ stats:
   pBSQDistribution: Mainnet BSQ distribution is slated for April 2019.
   pTotalTrades: >
     <p class='stat-label'>Total Trades</p>
-    <p class='stat-figure'>32,318</p>
+    <p class='stat-figure'>34,535</p>
     <p class='stat-note'>Through June 2019</p>
```
m52go: July
```suggestion
    <p class='stat-note'>Through July 2019</p>
```

huey735:
@wiz I'm afraid I don't know how to do that.

pt-PT/vision.md
```diff
@@ -4,6 +4,7 @@ title: Vision &lsaquo; Bisq - The decentralized Bitcoin exchange
 ref: vision
 lang: pt-PT
 language: Português (PT)
+outdated_translation: true
```
m52go, line 7: this is true...intentional?

huey735: Yes. I have purposefully not changed them yet as I want it to through Transifex and I wanted to text the alert tool.

wiz:
@huey735 no worries buddy but IMO this PR needs to be rejected - if you want to make a PR to "add french translations" then it should only contain that and not add unrelated new functionality. it kinda looks like you just commited your work tree with random other things in it, which is fine when you do that on your local tree, but you need to separate features when submitting PRs for merging so that we can cleanly review them one at a time.

huey735:
@m52go  @wiz Ok. I'll figure out how to separate the main three updates from this PR into three new PRs.

m52go:
Yeah I think wiz is right here.

https://github.com/bisq-network/bisq/blob/master/CONTRIBUTING.md#contributor-workflow

Note this part: "Pull requests should be focused on a single change."

Typically I tend to be more lax with this on the website repo since it's not the end of the world if there's a mistake (e.g., unlike the software where money can be lost, etc) but this PR is an example of why it's not ideal to bunch up different changes in the same pull request **for reviewers**.

@huey735 please make separate pull requests for the French translation, translation flag, etc.

huey735:
I think I got it.
https://github.com/bisq-network/bisq-website/pull/217
https://github.com/bisq-network/bisq-website/pull/218
https://github.com/bisq-network/bisq-website/pull/219

huey735:
Thank you very much @m52go and @wiz.

PR #218, opened 2019-08-26
Remove version variable from downloads page

huey735:


_includes/main_nav_tr.html
```diff
@@ -1,6 +1,13 @@
  <nav class="navbar navbar-expand-lg navbar-light bg-light fixed-top pl-md-5 pr-md-5 pl-sm-3 pr-0">
 
-    <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+   <div>
+     <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+
+     <!-- Outdated translation alert tool tag-->
+     {% if page.outdated_translation %}
+       <span class="navbar-text" style="color: orange;">Outdated translation</span>
```
wiz, line 8: It looks like this is doing more than simply removing a variable from the downloads page?

m52go:
`master` already includes the version variable removal, so I recommend removing that commit (9e49594441c612be932f1731579ba0c046d865b5) and renaming this PR something like "Add outdated translation flag". Technically that's the main purpose of this PR.

_includes/main_nav_tr.html
huey735: @wiz I decided to put this commit in this PR because it also affects the downloads pages. It felt wrong to remove the 'version' variables in one PR and then add 'outdated_translations' in another PR when they overlap files.

huey735 (commented):
Well that didn't have the result I had in mind. I guess I better make a new PR.

huey735:
New and better PR https://github.com/bisq-network/bisq-website/pull/220

PR #220, opened 2019-08-26
Add outdated_translation variable

huey735:
On @m52go's advice I changed the construction of the outdated
translation alert tool to be spefic to the page instead of
the language. I added a "outdated_translation" variable to
the front matter of the relevant translated pages. They'll
have the values 'true' or 'false' according to the state.

### DIFF
--- a/_includes/main_nav_tr.html
+++ b/_includes/main_nav_tr.html
@@ -1,8 +1,6 @@
  <nav class="navbar navbar-expand-lg navbar-light bg-light fixed-top pl-md-5 pr-md-5 pl-sm-3 pr-0">
 
-   <div>
-     <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
-   </div>
+    <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
 
     <!-- Language selector for mobile -->
     {% if page.lang %}
--- a/_includes/main_nav_tr.html
+++ b/_includes/main_nav_tr.html
@@ -2,11 +2,6 @@
 
    <div>
      <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
-
-     <!-- Outdated translation alert tool tag-->
-     {% if page.outdated_translation %}
-       <span class="navbar-text" style="color: orange;">Outdated translation</span>
-     {% endif %}
    </div>
 
     <!-- Language selector for mobile -->
--- a/_includes/main_nav_tr.html
+++ b/_includes/main_nav_tr.html
@@ -1,6 +1,13 @@
  <nav class="navbar navbar-expand-lg navbar-light bg-light fixed-top pl-md-5 pr-md-5 pl-sm-3 pr-0">
 
-    <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+   <div>
+     <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+
+     <!-- Outdated translation alert tool tag-->
+     {% if page.outdated_translation %}
+       <span class="navbar-text" style="color: orange;">Outdated translation</span>
+     {% endif %}
+   </div>
 
     <!-- Language selector for mobile -->
     {% if page.lang %}
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -7,7 +7,7 @@ layout: default
       <!-- Outdated translation alert tool tag-->
       <div class="col-md-12 order-md-1 text-md-left">
         {% if page.outdated_translation %}
-          <a class="outdatedTranslation" href="https://www.transifex.com/bisq/bisq-website/">Outdated translation - Help us to improve it!</a>
+          <p class="text-muted">⚠ This page contains some outdated translations. Help us improve <a href="https://www.transifex.com/bisq/bisq-website/">here</a>!</p>
         {% endif %}
       </div>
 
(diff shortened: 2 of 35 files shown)


m52go:
Thanks Huey -- much better PR. 

Another thing we need to make sure to test is how the site appears on mobile, since the whole site is built to be responsive.

Making the font smaller won't work...we need to figure something else out.

Screenshot from 2019-08-26 13-52-11

_includes/main_nav_tr.html
```diff
@@ -1,6 +1,13 @@
  <nav class="navbar navbar-expand-lg navbar-light bg-light fixed-top pl-md-5 pr-md-5 pl-sm-3 pr-0">
 
-    <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+   <div>
+     <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+
+     <!-- Outdated translation alert tool tag-->
+     {% if page.outdated_translation %}
+       <span class="navbar-text" style="color: orange;">Outdated translation</span>
```
erciccione: I suggest to add a link toTransifex, so that potential contributors know where to go if they want to help translate the page. The message could be something like `Outdated translation - [Help us to improve it!](https://www.transifex.com/bisq/bisq-website/)`.

Would be also cool to have the phrase translatable imho, so that when somebody is using another language the warning will be in their language.

huey735: @erciccione what do you mean by message?

erciccione: The content of the warning (that now is only `Outdated translation`)

huey735: Well, for that I think we'll have to move it from the navbar. There's not enough space there as it is.

erciccione: I don't think the navbar is the best place for that warning. I would put it in a dedicated snippet under the navbar. That way would be easier to see for the user, we would have more space so we can be more descriptive and styling form mobile would be easier to manage.

m52go:
Now I'm kind of debating the need for this mechanism in the first place. The website isn't updated that often. 

Major updates we should mostly know ahead of time and can plan to get the necessary translations ahead of time. 

Minor updates probably don't justify having a page-wide flag...it's not that helpful to the user anyway because they won't know what's up-to-date and what's not.

huey735:
@m52go I still support this tool or some version of it.
In my mind the flag is to let the user know that they should refer to the English version for the up to date info. It also works as an ad to translators.

_includes/main_nav_tr.html
huey735: @erciccione @m52go I've moved it from the nav bar to the page body. Maybe we can work more on the text to display something more inviting and to make sure they check out the English version.
outdatedtool

m52go:
I am in favor of showing a message to encourage translators to contribute updates. 

In that case, maybe it would be better to put a more subtle message toward the bottom (right above the site footer at the bottom of the page content) saying something to the effect "This translation is slightly out-of-date. Would you like to help?"

erciccione:
I wouldn't put it at the bottom. Users need to know if a document is up to date before they start reading it. That's better UX.

I would keep the warning at the top under the navbar, but i would make it more visible. Maybe i light green background?

_includes/main_nav_tr.html
huey735: @erciccione how about this, following the standard warning sign palette. Orange text on green background doesn't stand out that much.
outdated2

huey735: The black text however doesn't imply that there's an hyperlink to be clicked.

erciccione: I'm not great with colours so, keep that in mind :P, but i personally think this solution is better than simple text. What about having the hyperlink underlined? or underlined on hover?

huey735: @ericcione now it looks like this when hovered
outdated3

erciccione: I guess could be fine. Better wait @m52go's input on this

m52go:
Could we do something a little more elegant? I don't think the current proposal goes well with the site's aesthetic.

For example:

Screenshot from 2019-09-04 13-19-09

huey735:
@m52go Am I right in assuming that that warning sign at the beginning of the sentence is the unicide U+26A0 and not an image?

m52go:
Yes. This is exactly what I put in the if condition. Don't need any additional CSS either.

```
<p class="text-muted">⚠ This page contains some outdated translations. Help us improve <a href="https://www.transifex.com/bisq/bisq-website/">here</a>!</p>
```

huey735:
@m52go Cool, I had come down to something similar. Had span instead of p though.

_includes/main_nav_tr.html
```diff
@@ -1,6 +1,8 @@
  <nav class="navbar navbar-expand-lg navbar-light bg-light fixed-top pl-md-5 pr-md-5 pl-sm-3 pr-0">
 
-    <a class="navbar-brand" href="{{ site.url }}/{{ page.lang }}/"><img src="{{ site.url }}/images/bisq-logo.svg" height="30"/></a>
+   <div>
```
m52go: Any particular reason for adding this div element?

m52go (commented):
This looks ready to go, thanks! Just one small question (see comment).

m52go:
Notes:
* the `outdated_translation` variables need to be added to the French pages, but this doesn't need to be done immediately (in this commit) as having no value is also falsey (which results in desired behavior)
* with erciccione's #221 merged, we should be able to get rid of the flag on all stats pages

_includes/main_nav_tr.html
huey735: @m52go They were vestiges of when the warning tool was placed in the navbar.

m52go:
ACK

_layouts/page.html
```diff
@@ -4,6 +4,13 @@
 
 <div class="container pb-5 mb-5">
     <div class="row align-items-center">
+      <!-- Outdated translation alert tool tag-->
+      <div class="col-md-12 order-md-1 text-md-left">
+        {% if page.outdated_translation %}
+          <p class="text-muted">⚠ This page contains some outdated translations. Help us improve <a href="https://www.transifex.com/bisq/bisq-website/">here</a>!</p>
```
wiz, line 10: Don't make links like this:
<a href="/foo">Click here</a> for more information</a>
Make links like this:
<a href="/foo">More information</a>

PR #225, opened 2019-09-03
Update stats page for august

m52go:
Statistics really needs refactoring...looking forward to putting #221 in motion.


m52go:
Merging, as this is just a simple numbers update (as usual).

PR #226, opened 2019-09-04
Separate volume by fiat and altcoin

m52go:


erciccione (changes requested):
The stats.html page of the other languages need to to be updated too and would be better to put the text string in a variable, so that can be picked up easily by the other languages.

m52go:
Done. Thanks @erciccione, bad oversight on my part.

PR #228, opened 2019-09-05
Update software stats version and outdated_translation vars

m52go:
Mostly small text/front-matter corrections.

### DIFF
--- a/_data/de.yml
+++ b/_data/de.yml
@@ -305,7 +305,7 @@ stats:
   pTotalDate: Während des letzten Monats
   pDailyAverageTrade: Täglich durchschnittliche Trades
   pSoftwareDownloads: Software Downloads
-  pSoftwareDownloadsVersion: Für v1.1.2 (ungefähr)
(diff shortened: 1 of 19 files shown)


m52go:
Merging, as the corrections are minor but important.

erciccione:
I'm a bit worried by the fact that merge requests get merged so fast. I understand that these are minor corrections, but they are many and this repo depolys directly to the website. I would be more cautious as the bisq project grows bigger and i would wait at least for one review, or 24 hours, before merging :).

m52go:
@erciccione unlike with the software, my heuristic for merges in this repository are based on 1) subjectivity and/or 2) functional changes (i.e., stuff that may break the site). If the merge has even a modicum of subjectivity or functional effect, I always wait for at least 1 review before merging. 

This request was all typos/text which met neither condition above.

But this is mainly for expediency, as until recently there weren't any other regular contributors to this repository to give reviews. As that number increases, it may make sense to wait for feedback on requests like this too.

PR #231, opened 2019-09-05
Add missing "fr" and "fr-FR" locales in languages.yml

wiz:


m52go:
@huey735 FYI need to make these changes for new languages.

m52go:
utACK

PR #235, opened 2019-09-06
Avoid relative image URL confusion by referencing /images directly

wiz:


m52go:
NACK

Images at /dao for non-English pages don't work.

m52go:
ACK

PR #240, opened 2019-09-10
Replace broken relative screenshot URL with absolute URL

wiz:


m52go:
I agree with the proposed change, but just so that I'm clear on what this PR does, what's currently broken? Image URLs seem to resolve. 

I do see that French images are missing on the front page, but that requires a different fix and I assume isn't the purpose here.

m52go:
Never mind, for some reason it works on the live website but not locally. Technically it is wrong as it is.

m52go:
ACK

PR #242, opened 2019-09-10
Display name of languages and country flag in locale selector

wiz:
- Use full language name instead of two letter language code
- Use flag emoji instead of two letter country code
- Fix sort order of locale selector pull-down

Screen Shot 2019-09-11 at 4 34 16

### DIFF
--- a/_includes/main_nav.html
+++ b/_includes/main_nav.html
@@ -6,14 +6,14 @@
     {% if page.lang %}
     <div class="dropdown mobile-language-selector more-info-desktop">
       <button class="btn btn-sm dropdown-toggle" type="button" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-        {{ page.lang }}
+        {{ page.language }} {{ page.flag }}
       </button>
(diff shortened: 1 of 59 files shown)


m52go:
I agree this would be better, but responsiveness becomes an issue for languages with longer names.

Desktop

Screenshot from 2019-09-10 10-20-38

Mobile

Screenshot from 2019-09-10 10-19-31

wiz:
@m52go how's that

wiz:
Screen Shot 2019-09-10 at 23 37 27

m52go:
ACK

erciccione:
Flags are not a good idea when dealing with localizations. They can create confusion (a country flag doesn't rappresent a language) and in some cases even conflicts. I don't agree with this change, i would have NACKed it before but this PR, that changes *59 files*, was merged **6 hours after its creation** (which is still a very bad and dangerous workflow IMHO) and i wasn't able to see it before now.

Please don't rush with the merges. Beside the dangers, it's very hard to keep up with the development when PRs are merged after few hours.

m52go:
@erciccione let's handle the discussion of this commit in #243, and continue discussion about merges on Slack in #bisq-website.

PR #245, opened 2019-09-15
Fix fee schedule charts on dao blog post

m52go:
BTC fee numbers were switched, and there was a duplicate column for BTC taker fees.


wiz (approved):

wiz:
ACK, but this should probably be text table and not an image :)

Issue #246, opened 2019-09-15
Help creating a pull request:  Updating Fee info on the FAQ


ifarnung:
Current old outdated fee information:
"Trading fees
Offer makers pay 0.1% of the trade amount, and offer takers pay 0.3% of the trade amount. The minimum in both cases is 0.00005 BTC to avoid dust limits. Fees paid with BSQ can be as much as 90% lower."

Based on info received from m52go:
![fee-increases-cycle-4](https://user-images.githubusercontent.com/55181240/64929191-8058b000-d82b-11e9-9f6c-6f535a51d815.png)
Proposed update:

"Trading fees
Offer makers pay 0.2% of the trade amount, and offer takers pay 0.6% of the trade amount. The minimum in both cases is 0.00005 BTC to avoid dust limits. When fees are paid with BSQ, offer makers pay 0.13% of the trade amount, and offer takers pay 0.38% of the trade amount."

wiz:
Fixed in #245

m52go:
@wiz this issue is for the FAQ section of the website, whereas 245 was for the blog post. This issue is still relevant.

wiz:
Oh, right. How about we make a text table there then?

m52go:
Yeah a table makes sense. Then it can simply be:

```
Trading fees are payable in BTC or in BSQ.

<table>

The minimum trading fee is 0.00005 BTC to avoid dust limits. Trading fees are DAO parameters, and can fluctuate with DAO voting.
```

wiz:
Feel free to assign this to me, but I might not get to it for a few days

m52go:
I can knock it out...unless you would like to, @ifarnung. Thanks for raising this point regardless.

m52go:
Closing as complete.

Issue #244, opened 2019-09-15
Change localization process


erciccione:
This is something we briefly discussed in the past days. The way the website is translated now is a beginning, but not optimal. Some translations are in html files, other in yml and some sections need to be updated for each language. That makes working on translations hard and unmaintainable in the mid-long term.

As i suggested in past i think we should implement a better workflow. I think using the [jekyll-multiple-languages-plugin](https://github.com/kurtsson/jekyll-multiple-languages-plugin) to manage the localization of the webste is the way to go.

SInce this approach received some positive feedback in the past I already started to work on this. It will take some time because requires a complete change of the structure of the repo and of most of the files, but after, the maintainance of the repo will be easier and the structure will be consistent for all languages.

Feel free to assign this issue to me

wiz:
Does this plugin support country-specific locales?

m52go:
@erciccione this could be very good, I think, as long as:
1. you think the effort is worth it...it seems like translations are a bit of a mess to maintain regardless, and changing the localization process probably only makes sense if it's significantly better than what we have now.
2. this plugin can be made to accept locales, as we discussed in #243. I know you're not a fan of that direction, but that's currently the way we'd like to go.

erciccione:
The plugin doesn't care. It only changes the structure, we can use any language/locale we want. To clarify: My intention is not to revert or ignore the consensus reached in #243. A collective decision is law in an open source project, even if i was against it.

> you think the effort is worth it...it seems like translations are a bit of a mess to maintain regardless

If there is an interplanetary truth, is that translations are a mess to maintain :), but as i briefly explained, i think this plugin is really necessary. The present way is unmaintanable in the long run and touches too many files. It will only only get worse with every new language. This new way of doing things will make everything much easier for both maintainers and contributors.

m52go:
Great, I figured that, but just wanted to make sure. Assigned this issue to you :)

erciccione:
I'm not interested in working on this anymore. See https://github.com/bisq-network/bisq-website/pull/260#issuecomment-539960325.

erciccione:
On a second thought, i think it's better to reopen this issue. This repo is still in a very bad shape and if new languages are added with the current system, they will only make the repository heavier and more messed up than it already is. A refactor is very needed and somebody else could be willing to pick up this task.

PR #248, opened 2019-09-17
Update fee structure on faq page

m52go:
To address #246. We discussed making a new web page for fees, but that is probably best left for separate PR.

### DIFF
--- a/_data/en.yml
+++ b/_data/en.yml
@@ -215,71 +215,11 @@ faq:
     </ul> <strong>Altcoins</strong> <ul>
       <li>Blockchains: 1 day</li>
     </ul>
-  hsst: "How much does it cost to trade on Bisq?"
+  hTradeCost: "How much does it cost to trade on Bisq?"
   pTradeCost: >
     <p>To trade on Bisq, users pay (1) trading fees to Bisq and (2) mining fees to miners.</p>
 
-    <strong>Trading fees</strong>
-    <p>Trading fees are payable in BTC or BSQ. BSQ is colored bitcoin that Bisq uses to fund and govern itself. It's optional to use, but there are <a href="https://docs.bisq.network/getting-started-dao-traders.html" target="_blank">significant benefits</a>.</p>
-
-
-    <div id="faq-tables">
-
-        <p class="note">BTC trading fees, per 1 BTC traded:</p>
-
-        <table>
-            <thead>
-                <tr>
-                    <th colspan="1"></th>
-                    <th colspan="1">BTC fees</th>
-                    <th colspan="1">Fee as % of 1 BTC trade size</th>
-                </tr>
-            </thead>
-            <tbody>
-                <tr>
-                    <td colspan="1">Maker fee</td>
-                    <td colspan="1">0.002 BTC</td>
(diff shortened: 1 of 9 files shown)


m52go:
@ifarnung feel free to take a look at the preview too...let me know if you have any thoughts:

https://deploy-preview-248--bisq-website.netlify.com/faq/#cost-to-trade

erciccione (changes requested):
The changes to the faq page need to be added to the other languages as well.

m52go:
@erciccione thanks. I can only write in English, so I added the update to `strings_en.yml` and marked the other FAQ pages as outdated.

erciccione:
@m52go Wouldn't be better to have the english text in all the localized `faq.html` files as well? Otherwise i'm afraid we could lose these strings for the other languages (this problem will be solved by #244 tho, since a good amount of these duplicated files will be deleted)

m52go:
@erciccione how would they get lost? I thought the strings_en.yml file was the base file used for translating into other languages.

In any case, yes, seems like your refactoring is sorely needed. Something isn't right with this setup.

erciccione:
> I thought the strings_en.yml file was the base file used for translating into other languages.

Nope, that will be the situation once #244 will be implemented. Right now strings are split between `strings_en.yml` and html files and everytime something is changed the language files must be adapted. In this case the `faq.html` page of every language must be adapted. A further complication is that the strings of the language files are not in `[LANG]/faq.html` (like it is for the english strings), but in `[LANG].yml`. So, according to the current workflow the english strings are in `faq.html`, but the translated strings are variables in `[LANG]/faq.html` that will be pulled from `_data/[LANG].yml`.

Yes, it's a mess, that's why i agree #244 is badly needed :)

erciccione:
> Something isn't right with this setup

I think it was a mistake made during the conversion to the website to make it able to host translations. All strings should have been added to the `strings_en.yml`, but for some reasons most of them were left in their original file, which result in a very messed up structure and workflow.

m52go:
@erciccione I agree it's not ideal to have an answer that's not thorough (current answer with `outdated_translations` flag), but isn't it worse to have a thorough answer that no one can understand (new answer in English)?

@huey735 side note / curiosity: the [trading fees question on the FAQs page](https://bisq.network/faq/#cost-to-trade) doesn't seem to mention BSQ on any of the non-English pages. I wonder how that happened? BSQ has been a part of the English version for months now, presumably before any website translations started.

huey735:
@m52go I've just noticed that it's missing from `strings_en.yml` so it didn't make to Transifex, I'll get to it.

huey735:
@erciccione the strings I'm referencing are related to this PR. I'd be creating a new PR to update the strings_en.yml file.

m52go:
@erciccione @huey735 I've rebased this PR now that #256 is merged. 

What's the deal with this PR? Can it be merged?

I know the way it handles non-English languages isn't ideal, but that's not the point of this PR, and it's not a valid reason to hold it up. It uses the conventions of the current set-up, which is the best it can do.

erciccione:
@m52go I'm fine with this getting merged, it's not a big deal to adapt it to the new system.

PR #259, opened 2019-09-26
Update dispute resolution faq with trader chat

m52go:
Will make this answer more substantial when the related documentation is updated, but for now, this at least make the answer correct.


huey735 (approved):

m52go:
@huey735 thank you so much!

PR #263, opened 2019-10-02
Update stats with september numbers

m52go:
@huey735 @erciccione mind taking a quick look?


erciccione (approved):
Looks good to me.

utACK

PR #270, opened 2019-10-26
Don't use format=jscallback when calling Markets API

wiz:
Since the Bisq Markets API server will soon switch to a new backend that doesn't support `format=jscallback`, the Bisq website needs to convert the JS object into an array by itself, and the fix is quite simple.

### DIFF
--- a/js/markets.js
+++ b/js/markets.js
@@ -49,7 +49,11 @@ function buildTicker( pair ) {
 }
 
 
-function buildChartTitle( valueFinal, pair ) {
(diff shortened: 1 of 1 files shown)


wiz:
Hmm, found a bug on staging that didn't reproduce in my development environment. I'll have to figure it out and push another commit to fix this. https://deploy-preview-270--bisq-website.netlify.com/markets/

m52go:
ACK

PR #280, opened 2019-11-05
Change page titles: bisq is network not exchange

m52go:
Page titles across the site call Bisq "The decentralized Bitcoin exchange" which doesn't sound right...it's more of a network that serves as a means to exchange bitcoin.

Also "the" sounds pretentious, as if it's the only entity of its kind that will ever exist, so it's changed to "a".


RiccardoMasutti (approved):

devinbileck (approved):
utACK
I have also submitted a PR to update strings within the app.
https://github.com/bisq-network/bisq/pull/3681

freimair:
I like the "The". Marketing stuff?

@ManfredKarrer @ripcurlx @sqrrm

ripcurlx:
If this would be a company I would go for "The ..." as it is more common to set out big claims without the need to actually proof it 😉 . To reduce the troll attack vector I agree that a more humble approach might suite better for Bisq and changing away from a pure `exchange` to a `exchange network` communicates more what Bisq is as well. So from my side 👍 . But as this might be a bigger change also for the future I'd like more contributors to give their opinion as well before merging. e.g. @cbeams

m52go:
Since discussion is happening here, I'll copy my comment from https://github.com/bisq-network/bisq/pull/3681:

---

I'm not sure where/how "the exchange" wording originated, but I think it's pretentious and unusual.

Debian - _a_ free operating system (OS) for your computer ([link](https://www.debian.org/))
Qubes -  _a_ reasonably secure operating system ([link](https://www.qubes-os.org/))
Bitcoin - _an_ experimental digital currency ([link](https://github.com/bitcoin/bitcoin))

I'll wait to merge the website changes until this gets more consensus since it's a relatively meaningful change (i.e., definition of the project).

_data/languages.yml
```diff
@@ -2,7 +2,7 @@
   tag: "en"
   enabled: No
   accept_languages: ["en", "en-us", "en-au", "en-nz", "en-za", "en-bz", "en-tt"]
-  site_name: "Bisq - The decentralized Bitcoin exchange"
+  site_name: "Bisq - A decentralized Bitcoin exchange network"
```
cbeams: +1 for this change in general, i.e. using 'A' and 'network', for the same reasons @m52go laid out in his comments.

In addition, I suggest that we use lowercase 'bitcoin'. Per https://en.wikipedia.org/wiki/Bitcoin#cite_note-19:

> Some sources use Bitcoin, capitalized, to refer to the technology and network and bitcoin, lowercase, to refer to the unit of account.

While an argument could be made that Bisq is both a network for exchanging bitcoin (the currency) and that Bisq is an exchange network built on top of the Bitcoin protocol, the former meaning is what most readers of this description would naturally think about first, especially if they are not yet familiar with how Bisq works.

cbeams (changes requested):
ACK on the proposed changes plus a couple additional suggestions of my own.

_data/languages.yml
```diff
@@ -2,7 +2,7 @@
...
   site_desc: "Bisq is an open-source desktop application that allows you to buy and sell bitcoins in exchange for national currencies, or alternative cryptocurrencies."
```
cbeams, line 6: I realize that this text isn't part of the proposed pull request, but I suggest we change it to match the wording in the introduction of https://docs.bisq.network/exchange/whitepaper.html:

> Bisq is an open source peer-to-peer application that allows anyone to buy and sell Bitcoin in exchange to national currencies or alternative crypto currencies.

and actually, I'd suggest we tweak _that_ wording as follows:

> Bisq is an open source, peer-to-peer application that allows anyone to buy and sell ~~B~~bitcoin in exchange ~~to~~for national currencies ~~or~~and other ~~crypto currencies~~cryptocurrencies.

such that the final version reads like this:

> Bisq is an open source, peer-to-peer application that allows anyone to buy and sell bitcoin in exchange for national currencies and other cryptocurrencies.

A quick search will show that minor variations of the above wording have been used in many docs and articles around the web.

cbeams (changes requested):
I meant to add this comment to my review above, but it got lost in the process.

_data/languages.yml
m52go: Yes, fixed in https://github.com/bisq-network/bisq-website/pull/280/commits/6670f6130c6cf15c424f7b1ffc7b46565e48e193.

_data/languages.yml
m52go: Agreed. I've submitted [another PR for this](https://github.com/bisq-network/bisq-website/pull/303) since others have already reviewed this one.

RiccardoMasutti (approved):

m52go:
Merging, as remaining issues have been addressed, and accompanying PR in bisq-network/bisq is merged now too.

PR #292, opened 2019-11-21
Fix arch linux package link

positiveblue:
The current bisq.network page points to a bad repository which is not
uploaded anymore (bisq-git). This PR points to the right repo (bisq)

Changes made automatically executing

`grep -r --files-with-matches bisq-git | xargs sed -i "s/bisq-git/bisq/g"`

### DIFF
--- a/_includes/os_selector_options.html
+++ b/_includes/os_selector_options.html
@@ -46,7 +46,7 @@
   <img src="/images/icon-arch-linux-g.svg" class="os-icon os-icon-g" />
   <img src="/images/icon-arch-linux-w.svg" class="os-icon os-icon-w" /> Download for Arch Linux
   <div class="float-right">
-    <a href="https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=bisq-git">Arch User Repo</a></div>
+    <a href="https://aur.archlinux.org/packages/bisq/">Arch User Repo</a></div>
 </div>
 
 <div class="dropdown-item btn-icon">
--- a/_includes/os_selector_options_tr.html
+++ b/_includes/os_selector_options_tr.html
@@ -47,7 +47,7 @@
   <img src="/images/icon-arch-linux-g.svg" class="os-icon os-icon-g" />
   <img src="/images/icon-arch-linux-w.svg" class="os-icon os-icon-w" /> {{ item.liArch }}
   <div class="float-right">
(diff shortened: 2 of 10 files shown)


freimair:
actually, the package still exists and the link works. Just checked it again.

However, I agree that this is not perfect.
How about pointing to [https://aur.archlinux.org/packages/?K=bisq](https://aur.archlinux.org/packages/?O=0&K=bisq) instead?

positiveblue:
As @freimair says, the package still exist but it is not updated (v1.0.0). 

The two packages are maintained by the same person and I wonder if there is some historic reason to point to `bisq-git` which does not apply anymore.

We can point to `https://aur.archlinux.org/packages/?K=bisq` but it adds some complexity for new users. Basically they visit `bisq.network` for first time and they one to download and try the software. Not sue if showing them three options with only one working is the best experience for them.

IMO it is better if we directly redirect to the package that works, maybe `https://aur.archlinux.org/packages/bisq/`

m52go:
Which link do we want to use? I see the usability benefit of linking directly to the bisq page (i.e., not bisq-bin or bisq-git), but I assume most people in this situation will be able to figure out what they need if confronted with the search results page.

Either way, this PR needs to be updated with one of the links in order to be merged.

positiveblue:
PR has been updated pointing to the "correct" one (for most of the users) https://aur.archlinux.org/packages/bisq/

freimair:
the Arch User Repository (AUR) is the way of installing third-party stuff in arch linux. If someone uses Arch linux, they are familiar with the AUR and how it works. I, however, agree that we should point out that the "bisq-bin" contains a JDK, "bisq" does not.

ad bisq-git being outdated, well, it is a git package. Just build it and you get the latest master. That is how *-git packages behave.

ad only one working: which ones do not work?

positiveblue:
@freimair with not working I was talking about `bisq-git` which builds bisq v1.0


EDIT: After checking it in detail the confusion comes from `bisq-git` [packversion](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=bisq-git#n4) that is `1.0.0.r1.g60826a11e`. However, it is true that it builds the last version from master

For me, I would point to `bisq-bin` but whatever you prefer. If you think that this `bisq-git` is good enough we can leave it like this

m52go:
I'll merge this (after adding corresponding updates for a new language added in the interim) since there was rough consensus on the proposed link.

PR #294, opened 2019-11-25
Convert proper ja screenshots to svg

m52go:


wiz (approved):
Thanks for fixing this

PR #297, opened 2019-11-27
Fix es/dao/ button "Ver más"

luisantoniocrag:
This button will not work because it is linked to an id called #why and the section you should call has the id #porqué. I think it is better to keep everything in English so that these problems do not occur.

### DIFF
--- a/_data/dao_content_tr.yml
+++ b/_data/dao_content_tr.yml
@@ -80,7 +80,7 @@ es:
   target: "_blank"
   rel: "noopener"
   img: /es/images/DAO/dao_why.svg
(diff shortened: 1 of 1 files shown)


m52go:
ACK

m52go:
Thanks @luisantoniocrag, looks like translators didn't realize the `id` attributes weren't supposed to be translated. This is also a problem for the French translation, [which I just made a PR for](https://github.com/bisq-network/bisq-website/pull/299), if you don't mind giving it a quick review.

EDIT: nevermind, disregard, I didn't see you had already made a PR for it!

PR #298, opened 2019-11-27
Fix fr/dao/ button "En savoir plus"

luisantoniocrag:
fr-dao

This button will not work because it is linked to an id called #why and the section you should call has the id #Pourquoi. I think it is better to keep everything in English so that these problems do not occur.


m52go:
utACK

Issue #295, opened 2019-11-26
Fix fonts in front page screenshots


m52go:
Some browsers are rendering the SVG images on the front page with system/browser defaults, and it's ugly.

If (for whatever reason) the solution isn't straightforward, we should go back to PNG files.

EDIT: also address spacing, i.e. "BT C".

<img width="1247" alt="Screen Shot 2019-11-25 at 21 25 19" src="https://user-images.githubusercontent.com/735155/69681959-b6c26f00-10a7-11ea-9f6b-68b952fe4cb4.png">

m52go:
Closing as complete.

PR #278, opened 2019-11-04
[Proposal] Translate Bisq software images in the homepage #207

gilsanjuan:
I worked  14 .svg images from Bisq software, 7 images of Offer book screen ( bisq_screen0), and 7 images of Trades( bisq_screen1). Included EN, ES, DE, FR, PT, JP, CN translations

 I optimized bisq_screen0 to 59 KB, and bisq_screen2 to 38 KB.

At the end, I changed png image extension to svg for bisq_screen0 and bisq_screen1 on index.html, _includes/homepage_content.html and _includes/homepage_content_tr.html

PR #304, opened 2019-11-28
Fix weird bug with svg images on /index

m52go:
To address #295.

Issue seems to be OS-related, as it was introduced with #278, and only seems to occur on macOS (which explains why I never saw the issue when reviewing).

I don't deem it worthwhile to debug, so I've just switched all screenshots back to PNG. SVG files remain in the repository as "source" files in case the PNGs need to be updated later.

**Please review this PR on macOS or Windows, as the issue never occurred on Linux.**

https://deploy-preview-304--bisq-website.netlify.com/


cbeams (approved):
ACK. LGTM on macOS 10.14.6 + Firefox 70.0.1.

PR #306, opened 2019-12-03
Update stats page for november 2019

m52go:
This update switches from Slack user counts to Keybase user counts, and also reports downloads for v1.2.3 (formerly v1.1.7).


m52go:
Same update as always...will merge today if there are no objections.

PR #300, opened 2019-11-27
Add Schema.org JSON-LD Microdata for SEO (Articles)

RiccardoMasutti:
Following #290 and #291 

As specified on [https://developers.google.com/search/docs/data-types/article](https://developers.google.com/search/docs/data-types/article)

Final result:
Two non-carousel results for non-AMP web pages.

Carousel results for non-AMP web pages.

This will boost traffic and SEO for Bisq.network.

@m52go unfortunately, for `NewsArticle` we cannot use `Project` as `Publisher`. Google accepts `Organization` only. By the way, I specified `Bisq Decentralized Autonomous Organization` as `publisher.name`.

It does not mean in any way that Bisq is a "standard" company.
From now on it would be ideal to use this scheme for each article we publish and, if possible, add a unique image for each one.

PR #291, opened 2019-11-21
Add Schema.org JSON-LD Microdata for SEO (FAQ page)

RiccardoMasutti:
Following #290 

As specified on [https://developers.google.com/search/docs/data-types/faqpage](https://developers.google.com/search/docs/data-types/faqpage)

Final result:
image

PR #290, opened 2019-11-21
Add Schema.org JSON-LD Microdata for SEO (Homepage)

RiccardoMasutti:
We should add schema.org microdata to Bisq website for a better Search Engine Optimization as specified on [https://schema.org/SoftwareApplication](https://schema.org/SoftwareApplication) and [https://developers.google.com/search/docs/data-types/software-app?hl=en](https://developers.google.com/search/docs/data-types/software-app?hl=en).

Example of a Rich Snippet inside Google's SERP (search engine result page):
Google Rich Snippet

### What is Microdata?
Microdata (like RDFa and Microformats) is a form of semantic mark-up designed to describe elements on a web page e.g. review, person, event etc. This mark-up can be combined with typical HTML properties to dedlfine each item type through the use of associated attributes. For example, ‘Person’ has the properties name, url and title – attributes can be applied to HTML tags to describe each property.

### What is Schema.org?
Schema.org is a universally supported vocabulary extension by Google, Microsoft and Yahoo! for mark-up languages such as Microdata. It is designed to make the lives of webmasters easier, by offering one standardised mark-up understood by all the major search engines. Currently, Schema.org is compatible with Microdata, RDFa and JSON-LD.

### What is JSON-LD?
Based on the popular JSON format, the linked data format JSON-LD allows webmasters to define the context of the data contained through the use of types and properties. When combined with Schema.org, these properties follow a standardised mark-up supported by major search engines, and joins Microdata & RDFa as methods for integration. Unlike Microdata & RDFa, JSON-LD offers greater ease of implementation with all the necessary mark-up contained within inline <script> tags, instead of wrapping HTML properties. However, as elegant and lightweight that JSON-LD is, there are some potential road blocks. In some instances it’s just not practical to mark-up content, for example that on a larger scale, as the content would need to be effectively repeated within the script tags in order to validate. Also as the mark-up is invisible, the likelihood of marking up content that is not on the visible page increases, which is against search engine usage guidelines. It is for these reasons that Google in particular still favours Microdata & RDFa for marking up HTML content.

### Why use markups on Bisq website?
Marking up content on Bisq website can:
- Lead to the generation of rich snippets in search engine results like the one showed on top
- Enhance CTR (Click Through Rate) from the search result
- Provide greater information to search engines to improve their understanding of the content on Bisq website

Read more about the importance of Rich Snippets: [https://yoast.com/what-are-rich-snippets/](https://yoast.com/what-are-rich-snippets/)

PR #307, opened 2019-12-03
Add Schema.org JSON-LD Microdata for SEO (Articles)

RiccardoMasutti:
Following #290 and #291. Replace #300

As specified on [https://developers.google.com/search/docs/data-types/article](https://developers.google.com/search/docs/data-types/article)

Final result:
Two non-carousel results for non-AMP web pages.

Carousel results for non-AMP web pages.

This will boost traffic and SEO for Bisq.network.

@m52go unfortunately, for `NewsArticle` we cannot use `Project` as `Publisher`. Google accepts `Organization` only. By the way, I specified `Bisq Decentralized Autonomous Organization` as `publisher.name`.

It does not mean in any way that Bisq is a "standard" company.
From now on it would be ideal to use this scheme for each article we publish and, if possible, add a unique image for each one.

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -25,3 +25,28 @@ layout: page
   <br>
   <a href="/blog">« Back to Blog</a>
 </div>
+
+<script type="application/ld+json">
+    {
+        "@context": "https://schema.org",
+        "@type": "BlogPosting",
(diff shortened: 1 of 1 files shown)


m52go:
ACK

PR #313, opened 2019-12-12
Fix Opengraph tags

RiccardoMasutti:
The Open Graph protocol enables any web page to become a rich object in a social graph. For instance, this is used on Facebook to allow any web page to have the same functionality as any other object on Facebook.

This PR fixes `meta` tags, because they were wrong.


m52go:
utACK

PR #318, opened 2019-12-20
Add PGP SIGNATURE link next to download

### LINKED ISSUE #317: Download page (https://bisq.network/downloads/) has a bad link to "PGP Signatures."
@dscotese commented on [Wed Aug 14 2019](https://github.com/bisq-network/bisq/issues/3083)

The "PGP Signatures" entry on the download page (https://bisq.network/downloads/) ends up providing the Release Notes instead of the folder containing the signatures.  I found the file I needed (https://bisq.network/downloads/v1.1.5/Bisq-64bit-1.1.5.exe.asc) by simply guessing that adding the .asc to the end of the link to the .exe would get it.  The "PGP Signatures" link should point to a new page that has a link to each of the .asc files.

---

@battleofwizards commented on [Wed Aug 14 2019](https://github.com/bisq-network/bisq/issues/3083#issuecomment-521125514)

All signatures are present on the Release Notes page down below the text content. Seek assets.

The confusion comes from the fact that Release Notes are very long and the actual downloadable assets are buried deep.

I'm afraid GitHub does not offer linking to catalogs of released files. One can only link the page and individual files.

One solution would be to **link** the release notes file instead of putting the text directly on the page. This would be inline with Bitcoin Core: https://github.com/bitcoin/bitcoin/releases and would ensure the assets are easily visible by the users.

---

@dscotese commented on [Wed Aug 14 2019](https://github.com/bisq-network/bisq/issues/3083#issuecomment-521371292)

If updating the release notes is possible, the issue I had can be fixed by adding a line to the "How To Verify" section. I forked the repo but could not see where the release notes file was for me to make this change and then a pull request.  Here is how that section would look (with the line I'd add in **bold**):

How to verify signatures?
**Download the \*.asc file for the file you downloaded ("BINARY" below) from the bottom of this page**
`gpg --digest-algo SHA256 --verify BINARY{.asc*,}`
Replace BINARY with the file you downloaded (e.g. Bisq-1.1.5.dmg)

... and since I like to explain things, I'd also add this:
**gpg will search the download folder for the file you downloaded (it recognizes and remove the ".asc" to do this) and use the specified file to verify it using the signer's key.**

---

@freimair commented on [Thu Sep 19 2019](https://github.com/bisq-network/bisq/issues/3083#issuecomment-533054943)

that is confusing I agree. How about changing the download page and add links to the signature files right behind the binaries?
Screenshot from 2019-09-19 11-45-29

---

@stale[bot] commented on [Wed Dec 18 2019](https://github.com/bisq-network/bisq/issues/3083#issuecomment-566971610)

This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

RiccardoMasutti:
Closes #317


RiccardoMasutti:
@ripcurlx what about main menu?

m52go:
The links don't work! Just add `.asc` at the end of the download links. Your links are a bit different.

See:
https://deploy-preview-318--bisq-website.netlify.com/downloads/

Also I think it would be better to make the links lowercase, i.e. "PGP Signature" instead of "PGP SIGNATURE".

I think we can leave the main menus as they are, since there isn't much space there. Plus there's already a PGP signature link there anyway.

m52go:
@RiccardoMasutti updates with latest 2 commits get rid of merge conflicts and address other concerns...should be ready to merge.

PR #323, opened 2020-01-08
Align `The Bisq DAO` feature on mobile

RiccardoMasutti:
On Bisq.network homepage, right after `No Identity Verification`, `Decentralized`, `Safe`, `Private`, `Open` and `Easy to Use` feaures, there is `The Bisq DAO` (div id -> `#dao-callout`) which, on smartphone, is not aligned with the rest of the Features.

This PR fixes view the view on smartphones.

Screenshots from iPhone X:
**Before**
before

**After**
after

### DIFF
--- a/css/styles.scss
+++ b/css/styles.scss
@@ -1084,6 +1084,10 @@ faq page
     column-gap: 0;
   }
 
+  #dao-callout {
+    text-align: left;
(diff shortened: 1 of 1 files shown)


m52go:
ACK

PR #324, opened 2020-01-08
Replace Google Fonts with self hosted fonts

RiccardoMasutti:
There was a conversation a couple of days ago with other website devs and one of them asked me if having Google Search Console installed would be privacy invading or not.
I replied by saying that privacy is already broken by embedding Google Fonts.

> By using our APIs, you agree that Google can use submitted information in accordance with our privacy policies, such as http://www.google.com/privacypolicy.html

If we want to guarantee privacy to our users, it is necessary to remove Google Fonts and replace with self-hosted.
This PR replace Google Fonts with self hosted fonts on bisq.network website.

I have done several researches and spent many hours to make sure that this change does not detract from support for our users.
There are in fact different types of fonts. I decided to include only [.woff and .woff2](https://en.wikipedia.org/wiki/Web_Open_Font_Format) (less heavy, faster website loading) which are supported from:

- Firefox since version 3.6
- Google Chrome since version 6.0
- Internet Explorer since version 9
- Konqueror since KDE 4.4.1
- Microsoft Edge
- Opera since version 11.10
- Safari 5.1
- other WebKit-based browsers since WebKit build 528

After a careful analysis of the website visits (differentiated by browser version) through Matomo, I came to the conclusion that **this change affects ~ 0.01% of our users**.

In fact, it has emerged that, from mid-August to today, **only 114 users have used extremely obsolete browsers**. In particular:
- Internet Explorer <9 **112**
- Safari <5.1 **2**

_What happens if a user accesses the site with an unsupported browser?_
It simply does not load our fonts but uses those of the operating system.

If this PR is approved, I will do the same for the documentation.


m52go:
Thanks for tackling this! It looks good to me overall, but something isn't quite right with Japanese. 

Check the main page and DAO page as examples: font weight seems to be off, and some characters aren't rendered correctly in spite of the font being the same.

**Font weight example**, current then proposed:

Screenshot from 2020-01-08 15-56-10

Screenshot from 2020-01-08 15-56-15

**Font rendering example**, current then proposed (notice the character after "Bisq" text):

Screenshot from 2020-01-08 15-58-34

Screenshot from 2020-01-08 15-58-40

These examples are from the "Join the Community" section on the bottom of the front page.

wiz:
Japanese doesn't use the latin script bro, you want Japanese "regular" for m plus :wink:

_layouts/default.html
```diff
@@ -33,10 +33,7 @@
         <meta name="twitter:creator" content="@bisq_network">
         <link rel="canonical" href="{{ page.url }}">
         {% if page.lang == "ja" %}
-        <link href="https://fonts.googleapis.com/css?family=M+PLUS+Rounded+1c" rel="stylesheet">
-        {% else %}
```
wiz, line 37: Why are you removing this "else" statement? It breaks the Japanese font usage

css/jafont.scss
```diff
@@ -0,0 +1,14 @@
...
+# only Main files contain this front matter, not partials.
+---
+
+/* m-plus-rounded-1c-regular - latin */
+@font-face {
+  font-family: 'M PLUS Rounded 1c';
+  font-style: normal;
+  font-weight: 400;
+  src: local('M PLUS Rounded 1c'), local('MPLUSRounded1c-Regular'),
+       url('../fonts/m-plus-rounded-1c-v10-latin-regular.woff2') format('woff2'), /* Chrome 26+, Opera 23+, Firefox 39+ */
```
wiz:
```suggestion
       url('../fonts/m-plus-rounded-1c-v10-japanese-regular.woff2') format('woff2'), /* Chrome 26+, Opera 23+, Firefox 39+ */
```

css/jafont.scss
```diff
@@ -0,0 +1,14 @@
...
+       url('../fonts/m-plus-rounded-1c-v10-latin-regular.woff') format('woff'); /* Chrome 6+, Firefox 3.6+, IE 9+, Safari 5.1+ */
```
wiz:
```suggestion
       url('../fonts/m-plus-rounded-1c-v10-japanese-regular.woff') format('woff'), /* Chrome 6+, Firefox 3.6+, IE 9+, Safari 5.1+ */
```

css/jafont.scss
```diff
@@ -0,0 +1,14 @@
+---
+# this ensures Jekyll reads the file to be transformed into CSS later
+# only Main files contain this front matter, not partials.
+---
+
+/* m-plus-rounded-1c-regular - latin */
```
wiz:
```suggestion
/* m-plus-rounded-1c-regular - japanese */
```

wiz:
> Thanks for tackling this! It looks good to me overall, but something isn't quite right with Japanese.

I think it's because you're loading the IBM font on the Japanese site, you just need to use the "else" statement to only load the Japanese font (and not load the IBM font) on the Japanese site.

wiz:
@RiccardoMasutti if you add the japanese-latin font in one commit and remove it another, that font file will forever be included in the website repo, even tho we will never use it. can you squash those two commits together so the japanese-latin font is removed from the PR entirely?

RiccardoMasutti:
> @RiccardoMasutti if you add the japanese-latin font in one commit and remove it another, that font file will forever be included in the website repo, even tho we will never use it. can you squash those two commits together so the japanese-latin font is removed from the PR entirely?

Already done:
https://github.com/bisq-network/bisq-website/pull/324/commits/dcd9a5adec2252caa73466f93a9bd05421487a19
https://github.com/bisq-network/bisq-website/pull/324/commits/75cd0a9ad431945d2873f599c2da0edd32987ad9

wiz:
Right, but can you squash those commits into the commit that added the font file? Or simply squash all commits in this PR together will accomplish the same thing 😅

wiz (approved):
ACK but please squash commits before merging

m52go:
Squashed.

PR #328, opened 2020-01-10
Add Bisq Notification App source link

RiccardoMasutti:
In this way we give the possibility to directly access the source code and whoever wants can compile it out.

The next step would be to upload the Android version to F-Droid.


m52go:
To my knowledge, listing on F-Droid isn't possible because of Firebase.

Are we 100% sure it's possible to reproduce these builds? Would prefer to confirm this first. Otherwise we can just say something like "See source at...".

m52go:
Closing as stalled.

PR #330, opened 2020-01-10
Update ruby to v2.6.2

m52go:
Ruby 2.4 reaches EOL in March 2020.

It seems like Ruby does annual deprecations for their major versions, so [2.6 should be good through March 2022](https://www.ruby-lang.org/en/downloads/branches/). This PR moves to 2.6.2, which is the only version of Ruby 2.6 that Netlify seems to fully support at the moment.


m52go:
It seems Netlify is failing to build...will need to solve before this can be merged.

https://deploy-preview-330--bisq-website.netlify.com/

m52go:
Turns out 2.6.5 isn't fully supported by Netlify...2.6.2 is the only one that works.

RiccardoMasutti (approved):
ACK

m52go:
@ripcurlx FYI, just want to make sure you see this update, so you're not blindsided with an issue next time you update the site for a new release.

PR #335, opened 2020-01-27
Fix numbers on first 7 dao cycles

m52go:
Now using more precise means to calculate fees burned (most existing numbers aren't too far off).

Also made some silly mistakes at first while I was unfamiliar with various figures.


RiccardoMasutti (approved):
ACK

PR #337, opened 2020-02-08
Add L-BTC to markets currency selector

wiz:


m52go:
Update itself is fine and works, thanks. Typically we merge changes to the asset list with the corresponding release update, so I'm inclined to wait until 1.2.6 is actually released to merge this.

wiz:
Actually I think we need to upgrade Markets API, seednodes, website, etc. a few days before v1.2.6 is released so when users click on stuff it will work. I've already upgraded the Markets API to the release branch so this works fine: https://bisq.network/markets/?currency=l-btc_btc

ripcurlx:
Superseded by https://github.com/bisq-network/bisq-website/pull/341.

PR #339, opened 2020-02-10
Russian lang

FKrauss:
Added the two missing strings on the ` bisq-website/_data/languages.yml ` folder

Should be good to go now


ru/images/DAO/dao_benefits.svg
```diff
@@ -0,0 +1,195 @@
...
+        <tspan
+           sodipodi:role="line"
+           id="tspan4661"
+           x="472.00201"
+           y="2554.0569">Taxa de</tspan>
+        <tspan
+           sodipodi:role="line"
+           x="472.00201"
+           y="2572.8069"
+           id="tspan4665">negociação</tspan>
```
RiccardoMasutti: This is portuguese, not russian

RiccardoMasutti (changes requested):

RiccardoMasutti:
There are many strings translated incorrectly.

ru/images/DAO/dao_benefits.svg
FKrauss: true. nice catch. let me replace with the eng one then. because our Ru translators stopped replying

FKrauss:
Updated the images to the English version as pointed out by @RiccardoMasutti 

we don't have documented exactly where to fetch these strings from Transifex. They probably are translated there but without a way to verify it I rather keep in Eng.

We can plan on translating this at a later PR but I believe being in English is enough to publish

RiccardoMasutti (approved):
ACK

PR #340, opened 2020-02-10
Add Lazy Loading to images and iframes

RiccardoMasutti:
Replace #331

### Why native lazy-loading?

According to HTTPArchive, **images are the most requested asset type for most websites and usually take up more bandwidth than any other resource**. At the 90th percentile, sites send about 4.7 MB of images on desktop and mobile. That's a lot of cat pictures.

Embedded iframes also use a lot of data and can harm page performance. Only loading non-critical, below-the-fold images and iframes when the user is likely to see them **improves page load times, minimizes user bandwidth, and reduces memory usage**.

### SEO Benefits
Google has indicated site speed (and as a result, page speed) is one of the signals used by its algorithm to rank pages. And research has shown that Google might be specifically measuring time to first byte as when it considers page speed. In addition, a slow page speed means that search engines can crawl fewer pages using their allocated crawl budget, and this could negatively affect your indexation.

**Page speed is also important to user experience. Pages with a longer load time tend to have higher bounce rates and lower average time on page. Longer load times have also been shown to negatively affect conversions.**

Currently, there are two ways to defer the loading of off-screen images and iframes:

- Using the Intersection Observer API
- Using scroll, resize, or orientationchange event handlers

Either option can let developers include lazy-loading functionality, and many developers have built third-party libraries to provide abstractions that are even easier to use. With lazy-loading supported directly by the browser, however, there's no need for an external library. Native lazy loading also ensures that deferred loading of images and iframes still works even if JavaScript is disabled on the client.

Starting with Chrome 76, we are able to use the new `loading` attribute to lazy-load resources without the need to write custom lazy-loading code or use a separate JavaScript library.

Demo:
ezgif-7-61e4604beabd

Source: https://web.dev/native-lazy-loading/

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -6,9 +6,9 @@
 
         <div class="col-12 col-lg-4 mt-md-0 mt-3 pl-md-5 pl-sm-3 small px-sm-4 px-4">
           <h5 class="text-uppercase small grey">Bisq mobile notification App</h5>
-          <a target="_blank" rel="noopener" class="app-link" href="https://play.google.com/store/apps/details?id=com.joachimneumann.bisq"><img class="badge-img" alt="Get it on Google Play" src="/images/badge_google_play.svg"/></a>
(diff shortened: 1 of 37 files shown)


m52go (approved):
utACK

PR #342, opened 2020-02-12
Add q1 2020 update post

m52go:
Closes bisq-network/admin#21

PR #343, opened 2020-02-12
Add post on cycle 9 results

m52go:
Note: deploy is failing because the post in this pull request includes a `post_url` link to the [Q1 2020 post](https://github.com/bisq-network/bisq-website/pull/342), which hasn't been merged yet, so the link cannot be found and the build process throws an exception.

Just need to make sure PR 342 is merged first, and this PR should build without error.

EDIT: rebased so it builds successfully, now that 342 is merged.

### DIFF
--- a/_posts/2020-02-12-cycle-9-results.md
+++ b/_posts/2020-02-12-cycle-9-results.md
@@ -2,7 +2,7 @@
 layout: post
 title: "Bisq DAO Cycle 9: Results"
 author: Steve Jain
-excerpt: "Cycle 9 of the Bisq DAO ended at block 613 866 on January 21 2019. This post covers its results. <br><br>"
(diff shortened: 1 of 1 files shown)


cbeams (approved):
ACK. I'm very glad we do these posts.

Do you plan to cross post to the `@bisq-network/dao` team (as I believe we discussed)? I think we discussed cross-posting the full text, but perhaps it's enough to post the link (I don't want to create busy work for you transcribing the markdown). My concern is simply that people don't have a reliable mechanism for getting notified via RSS, and that many contributors will miss out on these updates.

m52go:
> Do you plan to cross post to the `@bisq-network/dao` team

Yes, can do it now just after merging. GitHub's discussions are formatted in Markdown too, so it's not a big deal. GitHub's flavor is only slightly different than whatever flavor it is Jekyll uses.

PR #348, opened 2020-03-02
Add <a> title attribute to languages

RiccardoMasutti:
Meta title tags are a major factor in helping search engines understand what pages are about. Title tags are used in three key places:
- search engine results pages (SERPs)
- web browsers
- social networks.


m52go:
NACK. See https://github.com/bisq-network/bisq-website/pull/350#issuecomment-594508001.

PR #352, opened 2020-03-09
Update stats page for february 2020

m52go:


m52go:
Merging since it's a straightforward numbers update.

PR #358, opened 2020-04-07
Update stat page for march 2020

m52go:
Routine monthly update.


m52go:
Codacy issue: "The numeric literal '449.350' will have at different value at runtime."

Not sure what could be changed...maybe remove trailing zero? Not worth figuring out, I think.

m52go:
Merging, as usual, since there's nothing out of the ordinary here.

Issue #356, opened 2020-03-13
Create friendly, interactive, simple getting-started page


m52go:
The existing getting-started guide is a long, static wall of text that only addresses 1 use case.

That page gets THOUSANDS of impressions a month, but I unscientifically guess that most of those impressions are wasted due to the nature of the medium and peoples' short attention spans.

I would like to replace links to this document with a more interactive and graphical getting-started page on the website that uses tweetstorms, videos, and graphics to convey the **bare essentials** that new users need to know in order to get going with Bisq—no matter what they want to use it for.

These new materials are in the process of being created. There's a [Bisq Distilled playlist on ](https://www.youtube.com/playlist?list=PLFH5SztL5cYOBvV3_TZJlFi14_65AbMwi)YouTube where you can see the first few videos, and there are some tweetstorms and graphics on Twitter that could go on this page as well.

More items are coming to complete [this initiative](https://github.com/bisq-network/growth/issues/185).

The page would be structured as a sequence of things to do, and there would be buttons for users to tailor the guide to their use case (e.g., download and verify on {Windows, macOS, Linux}...or set up payment method for {fiat, altcoin})...pick specific use case {USD, EUR, L-BTC, Monero}...etc.

I've assembled a very rough mockup of this page below. Please note that it will probably include some nicer copy and graphical/design elements...this is just to give a preliminary idea of what I have in mind.

Here's how I would like to use such a resource:
* redirect "get started" button on front page to it
  * integrate this call to action into a [revised front page layout](https://github.com/bisq-network/bisq-website/issues/355)
* redirect all major calls to action to it (e.g., link in twitter header)
* experiment with using it as a basis for workshops meetups around the world can use to onboard people to bisq

![Screenshot from 2020-03-13 13-13-29](https://user-images.githubusercontent.com/735155/76644923-dcd52780-654f-11ea-8c0e-4906b3c22eba.png)

m52go:
Should have a preview of this ready soon.

m52go:
Completed with #367.

PR #367, opened 2020-05-13
Add new getting-started page

m52go:
~For now, this is a prototype to address #356 I'm putting out mainly to gather feedback.~

Please check it out at https://deploy-preview-367--bisq-website.netlify.com/getting-started and let me know what you think.

Done:
- Basic layout and copy

Remaining:
- [x] Add this getting-started page to main nav (will need to move at least 1-3 other links elsewhere)
- [x] Switch redirects from current getting-started page in docs to this one
- [x] Make page responsive
- [x] Refine linked wiki pages

Will not do:
- Make page ready for translations (this will be done in a future PR after this page proves itself, and once the translations workflow is refined)

### DIFF
--- a/getting-started.html
+++ b/getting-started.html
@@ -38,10 +38,10 @@ en-only: true
 
         <div class="start-step">
             <div class="step-text">
-                <h2><span class="step-number">2</span>Set Up and Tour</h2>
+                <h2><span class="step-number">2</span>Back Up Keys, Write Down Seed</h2>
                 <p>With Bisq, you're in total control of your funds and your data. This means you retain unparalleled sovereignty, but it also means no one can help you if you lose something important—so it's critical that you do proper backups before using Bisq to trade.</p>
                 <p>We've also prepared a short tour of the Bisq interface so you can make the most of it.</p>
-                <p>At the very least, please be 100% sure to <a href="https://bisq.wiki/Backing_up_your_wallet" target="_blank">write down your seed words</a> and <a href="https://bisq.wiki/Backing_up_a_data_directory" target="_blank">back up your data directory</a>.</p>
+                <p>More details are on the wiki: <a href="https://bisq.wiki/Backing_up_your_wallet" target="_blank">writing down your seed words</a> and <a href="https://bisq.wiki/Backing_up_a_data_directory" target="_blank">backing up your data directory</a>.</p>
             </div>
             <div class="step-feature">
                 <div class="feature-selector"><a href="#" id="select-backing-up-video" class="backup-step-item active">Backing Up</a><a href="#" id="select-interface-tour-video" class="backup-step-item">Tour of Bisq</a></div>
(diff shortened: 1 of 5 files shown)


getting-started.html
```diff
@@ -0,0 +1,149 @@
...
+                </div>
+                <div class="responsive-youtube-container install-step-item" id="linux-install-video">
+                    <iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/dTfM4AsxNHY" frameborder="0" allow="accelerometer; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+                </div>
+            </div>
+        </div>
+
+        <div class="start-step">
+            <div class="step-text">
+                <h2>Set Up and Tour</h2>
```
cd2357: I would rename this section to "Backup your Bisq seed" since that's probably the key action point here

getting-started.html
```diff
@@ -0,0 +1,149 @@
...
+                <p>With Bisq, you're in total control of your funds and your data. This means you retain unparalleled sovereignty, but it also means no one can help you if you lose something important—so it's critical that you do proper backups before using Bisq to trade.</p>
+                <p>We've also prepared a short tour of the Bisq interface so you can make the most of it.</p>
+                <p>At the very least, please be 100% sure to <a href="https://bisq.wiki/Backing_up_your_wallet" target="_blank">write down your seed words</a> and <a href="https://bisq.wiki/Backing_up_a_data_directory" target="_blank">back up your data directory</a>.</p>
```
cd2357: "At the very least, please be 100% sure to backup .." => "Follow these simple steps (+link to wiki) to backup.."

The first version sounds like its kind of optional and we're kind of asking them to maybe consider doing it, at the very least maybe somehow think about it a bit.

Second version: hey go here and do it :) More imperative, people will see it as less optional, probably more will do it.

cd2357 (commented):
ACK / LGTM

A few small suggestions:
* see comments above for some suggested rewording
* might make it easier for people to follow along if the steps (so the titles of the sections) are numbered:

1) Download and Install Bisq
2) ...
3) ...

FKrauss:
Functionality wise it looks all good. I tested all buttons. The one weird thing I found (but can be improved later) is that if you start a video, then switch the filter (from Windows to Mac) the video doesn't stop and you can potentially have more than one video playing on screen. 


Design/conversion wise I think we need to do two things:

- Give the filters less highlight - have some quieter color to not take attention away
- put a big fat "Download now"  button on the first section below the text, see here:
image

Maybe even the same button we have in the home page.

cbeams:
I haven't watched each of the videos in detail, but I just want say that I love how this new page strikes a balance between concision and completeness. It's visually attractive and and invitingly short at first glance, making me feel like I can get through it, and rather than having to slog through paragraphs of prose instruction (however well written), I get the context I need in a brief blurb, and can then follow along in the video, which, due to the nature of the medium, I'm confident by default that I'll be able to do.  Nice!

Related, this makes me think we should do away with the https://bisq.wiki/Get_started article altogether and link to this one as the canonical resource.

m52go:
Looks like this is good to go. Just need some wiki pages to be updated, but the wiki was down until about a day ago because of some significant spamming issues. I'll work on the new articles and then merge this when appropriate. 

Thanks everyone for the feedback!

PR #373, opened 2020-06-23
Add post for cycle 13 results

m52go:


wiz (approved):
ACK - didn't verify the numbers so taking your word on it

PR #375, opened 2020-07-13
Update stats for june 2020

m52go:


m52go:
Usual update, albeit a bit late this month. Merging.

PR #378, opened 2020-07-22
Dark mode

pedromvpg:
Add darkmode functionality and new client screen images.


cbeams:
This looks great, wow. A few notes:

1. Consider defaulting to dark mode when the OS is in dark mode; see https://stackoverflow.com/a/52986538 for the CSS.
2. Swap the homepage screenshot to Bisq in dark mode when website is in dark mode
3. Text in the input field for the Markets page is hard to see note how the 'eur' text is dark-on-dark below:

image

cbeams:
Defaulting to dark mode seems to work well (I'm looking at https://deploy-preview-378--bisq-website.netlify.app/). One nit is that when dark mode is active, the menu toggle reads 'Dark', and when light mode is active, it reads 'Light'. Shouldn't this be the other way around? i.e. an affordance should advertise the state change that will occur if you make use of it (click it), as opposed to advertising the current state. Also, a common idiom I've seen is the use of a crescent moon symbol in the upper right for toggling dark / light mode. It takes up less real estate and is somehow more intuitive (to me) than just a textual 'light/dark' link.

m52go:
Just want to say I've been watching this PR -- it looks amazing! 

I see the markets dropdown text has been fixed, but the screenshots on the front page are still in light mode. Also I have agree with cbeams' comment above about flipping the dark mode toggle action and considering symbols instead of text for the label.

_layouts/default.html
```diff
@@ -120,6 +121,21 @@
     </head>
 
     <body class="home page-template-default page page-id-6 custom-background">
+    <script>
+      if (Cookies.get('darkmode') === undefined && window.matchMedia && window.matchMedia('(prefers-color-scheme: light)').matches) {
```
m52go:
```suggestion
      if (Cookies.get('darkmode') === undefined && window.matchMedia('(prefers-color-scheme: light)').matches) {
```

_layouts/default.html
```diff
@@ -120,6 +121,21 @@
...
+        console.log('no cookie');
```
m52go:
```suggestion

```

_layouts/default.html
```diff
@@ -120,6 +121,21 @@
...
+        Cookies.set("darkmode", 0);
+      }else if(Cookies.get('darkmode') === undefined && window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
```
m52go:
```suggestion
      }else if(Cookies.get('darkmode') === undefined && window.matchMedia('(prefers-color-scheme: dark)').matches) {
```

m52go (changes requested):
I had another look and everything seems to work fine. It looks so good (!).

I've attached a couple of nitpicks in the JavaScript...there's a stray `console.out()` still there from debugging and an unused condition in 2 of the if statements.

That's it from my end. @pedromvpg I will ACK once you make those changes, and merge a short while afterward if there's no other feedback.

m52go (approved):
ACK

PR #379, opened 2020-07-23
Add report for cycle 14

m52go:


MwithM (approved):
I did not check "proof-of-burn" numbers, but looks good.

PR #381, opened 2020-07-30
Separate javascript into page-specific files and remove unused code

m52go:
I wanted to add an anchor link to the front page, but FAQ page JavaScript was getting in the way. 

As it turned out, JavaScript code for the home and FAQ pages was being loaded for _every_ page. This also made load times unnecessarily slower for other pages.

This PR puts JavaScript for the home page in a separate home.js file that is only loaded when the home page is loaded. Likewise for the FAQ page.

Additionally, it removes lingering code from when Google Analytics was used.

Review/test:
https://deploy-preview-381--bisq-website.netlify.app/

## Review Guidance

In this case, there are just 2 commits, so it's worth looking at them separately.
- https://github.com/bisq-network/bisq-website/pull/381/commits/288eef800ee59b97a1b751c122ebc0a050ab96c8 just removes the `sendAnalytic()` function and both lines that call it
  - there isn't much to test here, aside from maybe inspecting the rest of the code base for any lingering mentions of `sendAnalytic()`...since these proposed changes remove that function, that method should not be used anywhere else.
- https://github.com/bisq-network/bisq-website/pull/381/commits/0605baadbdb6e0fb6d52a0672c1406d560719150 separates JavaScript in `/js/scripts.js` into 2 separate files called `/js/home.js` and `/js/faq.js`.
  - one way to test this is to make sure that all dynamic elements on the home page and faq page still work (e.g., accordion expanders on faq page, hover effect on getting-started section of front page, etc)
  - another good check could involve diffing the _site/ folder before and after these proposed changes are put in place. Since nothing visual should have changed, the generated website should be almost identical. To do this, just build the site without proposed changes, build the site with proposed changes, and then compare the result:
    - run `git checkout master`, `rm -rf _site`, and then `bundle exec jekyll build`. 
    - move the resulting `_site` folder somewhere else. 
    - then run `git checkout move-js`, `rm -rf _site`, and then `bundle exec jekyll build`. 
    - compare the resulting `_site` folder with the other one with a diffing tool of your choice (e.g. Meld). 

You should see `js/scripts.js` missing on every page, and `js/home.js` and `js/faq.js` added to all home pages and all faq pages...but everything else should be the same.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -66,7 +66,7 @@
         <script defer src="/js/jquery-3.3.1.min.js"></script>
         <script defer src="/js/popper.min.js"></script>
         <script defer src="/js/bootstrap.min.js"></script>
-        <script defer src="/js/scripts.js"></script>
+        <script defer src="/js/site.js"></script>
 
 {% unless page.layout == "post" %}
         <!-- Schema.org JSON-LD Markup for Search Engine Optimization -->
--- a/_layouts/page_faq.html
+++ b/_layouts/page_faq.html
@@ -1,6 +1,9 @@
 ---
 layout: page
 ---
+
+<script defer src="/js/faq.js"></script>
+
 {% assign item = site.data[page.lang].faq %}
 <div itemscope itemtype="https://schema.org/FAQPage">
   <h1>{{ item.hFAQ }}</h1>
--- a/_layouts/page_home.html
+++ b/_layouts/page_home.html
@@ -2,6 +2,8 @@
 layout: default
 ---
 
+<script defer src="/js/home.js"></script>
+
 {% assign item = site.data[page.lang].homepage_content %}
 <div class="container fadeIn">
         <main class="bd-masthead" id="content" role="main">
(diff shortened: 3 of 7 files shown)


Bayernatoor:
288eef8:

1.  Reviewed the code base - found no other mention of the `sendAnalytic()` function.
____

0605baa: 

1. Dynamic elements checked and working on local page and netlify, expanders and hover effect. (did notice that i couldn't switch to light mode, only dark mode was available).

2. Compared `_site` dir from master against proposed branch (correct term? ). `script.js` has been removed and replaced and added to the `js` directory `home.js` and `faq.js`

Looks good to me ... let me know if i went about it wrong at any point :) 

Not sure whether the light mode/dark mode not working is an issue.

m52go:
> (did notice that i couldn't switch to light mode, only dark mode was available).

This was **crucial**. Thank you for finding this. Yes I completely removed the code for the toggling dark mode when deleting `scripts.js`...the code for the toggle was not commented and kind of hidden at the bottom, so I didn't see it.

I've added it back now and rebased the branch so it's included in the last commit (https://github.com/bisq-network/bisq-website/pull/381/commits/ec7f8eaad7f9d3413e6499211fd3c07bc78c8368).

Issue #355, opened 2020-01-31
Experiment on changing the Bisq landing page


FKrauss:
## Learning Goal/Objective
> _- What do we need to learn?_
We currently have a very stable website conversion rate(visits with at least 1 download of the app). It hovers around 15% . The question is: is that good? is that bad? can we improve it?
![conversion rate](https://user-images.githubusercontent.com/4820683/73540399-1a627300-4430-11ea-9153-9014822f315d.png)

Also, looking at our "landing page" from a performance marketing/conversion perspective, I would say it is confusing (above the fold, there are 2 CTAs (both very shy), and the button to GET STARTED actually takes you to read the manual, not to download the app.

We don't have the funnel tracking completely there but we can add it. But, the biggest outlink from our website is the getting started section of the Docs website:
![2020-01-31 13_47_10-Bisq - From 2019-09-10 to 2020-01-31 - Web Analytics Reports - Matomo](https://user-images.githubusercontent.com/4820683/73540493-5ac1f100-4430-11ea-92ea-8bbc2169e2ae.png)

So the main problem question here is: **can we improve the conversion rate of our website?** i.e. get more visitors to download the app

Along those lines I did a small research on how other free downloadable software "companies" have their landing pages. You can find it [here](https://docs.google.com/presentation/d/1gsvt-d3t9nuPiM9GfV9njXGLIJOxnwOijH3wy_k8ims/edit#slide=id.p1)

## Proposal
What if we redesign  the home page in specific languages (or the whole site to experiment) to look closer to what other downloadable software guys are doing and see if conversion rates grow?

We could potentially do that in some languages of the site and track the conversion rate on Matomo with little effort

## Action Items

1. Figure out how to alter, design wise, pages on the website. Who knows, apart from people who initially built the website?
2. Figure out if we can isolate this change to some languages?
3. propose a "design" and content for the landing page
4. evaluate the experiment in matomo
5. decide whether or not to keep the change (and make the appropriate code adaptations for it)

-------
Personal comment, I need help with 1 and 2. With 3 I can start the process and gather feedback. 4 I can do it completely. 5 we need to discuss on Keybase or here

m52go:
I know the implementation of this issue didn't quite go as initially planned, but it was at least partially addressed with #382 -- calls to action are much more clear now. Of course, conversion rate is an ongoing concern, so we should continually monitor it and determine ways to improve it.

[...]

PR #382, opened 2020-07-31
Minimal design tweaks

pedromvpg:
- CTAs more prominent 
- Make navigation flat
- Add video to homepage

Screen Shot 2020-07-30 at 10 21 27 PM

at


m52go:
@pedromvpg loving the new aesthetic!

Feedback on the current front page call-to-action buttons motivated #355. This PR helps address that issue by making the download dropdown more prominent, but I wonder if we can do even better.

1. What do you think about making the dropdown into a button? Simpler, clearer call to action. Could mitigate lack of options (or wrong OS guess) with a simple link to downloads page.

Idea (if better-suited for a different PR, just let me know):

cta-suggestion

2. Personally the grey get-started button looks discouraging to me, or at least not encouraging.

What do you think about a different color?

cta-suggestion-2

3. The grey buttons look particularly bland in the mobile view...I think at least the Downloads button should be green, just like in desktop view, and the others ones possibly adjusted too.

Screenshot from 2020-08-04 11-27-51

pedromvpg:
Screen Shot 2020-08-05 at 9 31 27 PM
Screen Shot 2020-08-05 at 9 31 30 PM
Screen Shot 2020-08-05 at 9 32 01 PM
Screen Shot 2020-08-05 at 9 31 51 PM

PR #387, opened 2020-08-10
Simplify main_nav yml

m52go:
This PR simplifies the YML markup for ` _includes/main_nav.html`.

Previously, the `main_nav` markup was broken into parts and included URLs, which was unnecessary and confusing for translators. These changes condense the `main_nav` markup into a succinct list.

## Review Guidance

A thorough visual review of the main navigation menu should suffice to review this pull request. Visit https://deploy-preview-387--bisq-website.netlify.app/ and make sure the main navigation menu at the top matches what's on https://bisq.network for all 9 languages—desktop and mobile. There should be no missing links, etc. Ideally no colliding text either (e.g., from text that's too long).

### DIFF
--- a/_data/de.yml
+++ b/_data/de.yml
@@ -1,29 +1,14 @@
 main_nav:
-  liDownloads: Downloads
-  liLearnMore: Mehr erfahren
-  liDocumentation: Dokumentation
-  liBlog: Blog
-  liCommunity: Community
-  buttonDownloads: Downloads
-  links:
-    - title: Erste Schritte
-      url: /getting-started/
-    - title: Bisq DAO
-      url: /de/dao/
-    - title: Statistiken
-      url: /de/stats/
-    - title: FAQ
-      url: /de/faq/
-    - title: Märkte
-      url: /de/markets/
-  mobile_links:
-    - title: Community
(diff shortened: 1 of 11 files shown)


Bayernatoor:
Desktop verison checks out except - see screenshots and note below.
Japanese text showing up wrong on proposed branch
japanese
Main website:
Main-Japanese


Mobile tested on Android 8 and IOS 12 

Text looks good, navigation menu looks good.

------------

Desktop and mobile, regardless of language the Getting Started link always redirects the user to the English "Getting Started" page. Perhaps, this hasn't been translated yet?

m52go:
> Japanese text showing up wrong on proposed branch

Thanks, good catch. The translation seems to be correct, but the text is too long for the container. The previous translation was shorter, which is why this wasn't happening before. I've reverted this update with https://github.com/bisq-network/bisq-website/pull/387/commits/b0bf8e83cedae2e91dd1c2f23a6fb7beb3478221.

Likelihood of this happening in the future should decrease as a result of Transifex notes indicating maximum length for strings like this.

> "Getting Started" page. Perhaps, this hasn't been translated yet?

Yup, that's correct.

PR #388, opened 2020-08-11
Simplify dao yml

m52go:
Similar in spirit to #387, this PR simplifies the YML markup for `_layouts/page_dao.html` (the DAO page).

Previously, this markup included HTML markup and other artifacts that didn't belong. These suggested changes separate concerns better (keep text-to-be-translated in YML and HTML markup in HTML) and update DAO page copy.

## Review Guidance

As with #387, a thorough visual review should be sufficient here too. But since DAO page content is identical on desktop and mobile, there's no need to evaluate mobile rendering separately. Just browse the DAO pages throughout all 9 languages and ensure the text is all there. 

Text won't be identical since there were some translation updates...main things to look out for are missing paragraphs, obviously wrong text, broken links, etc.


Bayernatoor:
DAO page, visual review of all 9 languages: 

Formatting for Japanese text seems incorrect vs English version (master branch also appears this way)

image

image

Everything else looks good! 

663990a  - looks much cleaner then master.

a305b2e - text looks good and is a better "why?" then previously. it's appearing on local (en) branch fix-dao-yml

m52go:
> Formatting for Japanese text seems incorrect vs English version

Nice—HTML tag was missing. Thanks!!

PR #391, opened 2020-08-27
Revise diagrams and outline typography

pedromvpg:
New illustrations with outlined text for user without the font file and other small revisions to the diagrams

a


m52go:
New images look great!

I just rebased the commit in this PR on current master and updated DAO images for other languages too. Will need to update the text for other languages by hand since [my script](https://github.com/bisq-network/bisq-website/blob/cbc3197eda8a54fe6b0b8184bf30c2b9b0770a05/.tx/update-svg-text.js) doesn't seem to work.

m52go:
Carved out the perfect time to work on this on a recent flight...but then realized that none of the new SVGs have text in them, so they cannot be translated.

As a result, I think most reasonable thing to do now is just keep all pictures in English.

So I'm going to merge this PR as-is and make another PR to remove language-specific DAO images.

PR #393, opened 2020-09-03
Remove vision page

m52go:
Vision page reads like it was written before Bisq was built. Now that it's built and been around for a while, it may not be needed any more. FAQ page is where most people go to get their initial questions answered.

[Longer narrative-style introduction is on the wiki](https://bisq.wiki/Introduction), but I'm open to suggestions. Maybe an abridged version of the vision could be added as a FAQ item (?).


Bayernatoor:
Ya doesn't seem necessary anymore. Perhaps including this section or even just the lower part of the section pictured in the screenshot in the FAQ **(Bisq is built for those who)**

Screenshot from 2020-09-23 08-36-05

m52go:
Rebased on latest master branch, merging.

PR #394, opened 2020-09-03
Remove community page

m52go:
The community page serves no purpose that the community section on the front page doesn't already achieve.

This PR makes that section linkable so https://bisq.network/community links resolve.


Bayernatoor:
ACK

PR #396, opened 2020-09-03
Add new post: "Trends in Trading Activity, August 2020"

m52go:
Written to inform users about trends in trading (i.e., pricing premiums), but also intended to kick off discussions on a new growth initiative to get to 100,000 trades.

Preview link:
https://deploy-preview-396--bisq-website.netlify.app/blog/trading-trends-august-2020/


pingiun:
I like it! Very informative

MwithM:
That's a nice and helpful post to get a quick view of how could traders expect when trading at Bisq.  
All the info is displayed from BTC buyer's view, which in my opinion could discourage pointing the other, better looking point of view; sellers will take profit from selling at Bisq. Even if a seller doesn't care about privacy, to use Bisq offers a big economical incentive.
Also, I miss some mention to how much costs could affect this data: as mining fees are not % based but a fixed cost, it's normal that smaller offers require higher premiums.
There's different costs between making vs taking an offer: Making an offer is already encouraged by Bisq trading fees, as makers pay 0.05% and takers pay 0.35% for each trade (paid in BSQ). If we add this to the premiums paid by takers, making an offer is even better.

Bayernatoor:
Great article!

Good point MwithM ...I think for the most part (opinion) Bisq users are often motivated to BUY Bitcoin on Bisq to avoid the KYC. Sellers may not necessarily care that much but they DO care about the possibility of getting a better rate! Which is what the data seems to indicate.

I assume this type of article may be released every quarter or so?

m52go:
Thanks folks. Excellent feedback @MwithM! Important points. I added all of them, as well as a new TLDR section for impatient people.

> I assume this type of article may be released every quarter or so?

Yeah that would be great, or at least semi-annually.

MwithM (approved):
Thx for collecting my feedback.

Bayernatoor:
Ah ya good call on a TLDR!

PR #406, opened 2020-10-31
Update /markets frontend for changes in the Bisq Markets API v1.4

wiz:
In the bisq.markets `/api/trades` endpoint, the `direction` parameter was removed, so this PR removes the "ACTION" column from the "Trade History" table, which is now broken.

スクリーンショット 2020-10-31 11 33 52

### DIFF
--- a/js/markets.js
+++ b/js/markets.js
@@ -123,7 +123,6 @@ function getTrades( pair ) {
 
             $( '<tr>' ).append(
                 $( '<th>' ).text( 'Date' ),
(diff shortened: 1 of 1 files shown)


wiz:
Tested OK with both fiat and crypto markets:
https://deploy-preview-406--bisq-website.netlify.app/markets/?currency=bsq_btc
https://deploy-preview-406--bisq-website.netlify.app/markets/?currency=btc_usd

m52go:
ACK

PR #413, opened 2020-12-30
Add report for cycle 18

m52go:
Sorry for this being so late -- my typical workflow for aggregating and interpreting DAO figures stopped working, and adjusting it and getting this post done took longer than it should have.


MwithM (approved):
ack

PR #420, opened 2021-01-28
Update active markets using latest data from Bisq Markets API

wiz:
It seems a few shitcoins fell out of the active list, and a few new fiat markets have trades, so we should update the website list


m52go:
utACK

PR #432, opened 2021-04-29
Remove faq page

m52go:
It's now at https://bisq.wiki/Frequently_asked_questions.

This FAQ page was notorious for being out-of-date and stale. The wiki is much easier to maintain, and translation mechanism is built-in (can be activated whenever needed).

Although the Bisq software links to this page, this pull request includes a redirect to the wiki, so merging before those links are changed shouldn't create any issues.

PR #439, opened 2021-07-28
Remove markets page

m52go:
Remove markets pages, and link /markets to https://bisq.markets.

This new markets page has been live for a while, and fulfills exactly the same purpose as the current /markets page, but does it better (more performant, better design, etc).


m52go:
Merging for expediency. This PR is conceptually very similar to #432.

PR #444, opened 2021-09-26
Temporarily remove monitor.bisq.network links

xanoni:
Commented out the monitor for now given it's been broken for a year or
so, which could be wrongly interpreted as "Bisq itself has problems".

Files affected: _data/*.yml

See:
- https://github.com/bisq-network/bisq/issues/5297#issuecomment-927376913
- https://github.com/bisq-network/bisq/discussions/5340
- https://github.com/bisq-network/compensation/issues/685#issuecomment-927377283

### DIFF
--- a/_layouts/page_stats.html
+++ b/_layouts/page_stats.html
@@ -66,4 +66,5 @@ layout: page
 
 
 <br>
-{{ item.pNetworkLoad }}
+<!--Below line hidden temporarily until monitor site gets fixed:
+{{ item.pNetworkLoad }}-->


ripcurlx:
Yes, probably a good idea as long it is not fixed. I still use parts of the KPIs, but no need to link to it publicly as long as not everything is up and running.

m52go:
@xanoni agree, thanks for the suggestion. However, a better way to implement this would be to comment out the last line in `_layouts/page_stats.html`. Could you please make this change and rebase / resubmit?

xanoni:
> @xanoni agree, thanks for the suggestion. However, a better way to implement this would be to comment out the last line in `_layouts/page_stats.html`. Could you please make this change and rebase / resubmit?

Yes, that's definitely cleaner. I didn't realize it was that easy.

m52go:
utACK

xanoni:
Thx!

PR #472, opened 2023-02-21
Fix link to API

SmallAndSoft:


gabernard (approved):
utACK

Thank you for fixing the link!

PR #474, opened 2023-07-13
Updates for v1.9.12 release

alejandrogarcia83:


gabernard (approved):
utACK

gabernard:
The commit message has v1.9.**10** instead of 12 but the change is correct.

PR #481, opened 2024-03-13
added meta test for deployment test

salokiam:
The website should have a new META Tag now-
     <meta name="test" content="deleteme">

This can be used to test if the deployment has succeeded without change the website visually.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -12,7 +12,8 @@
         <meta charset="UTF-8"/>
         <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0">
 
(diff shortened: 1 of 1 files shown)


cbeams:
Thanks @salokiam!

cbeams:
I've confirmed that the new meta tag shows up as expected. @salokiam could you `git revert` this commit and push another PR so as to delete this test change?

Note for future reference, please form commit messages according to the conventions at https://cbea.ms/git-commit/#seven-rules. These conventions are used pretty consistently throughout bisq-network repos. Case in point here is the subject line. Under these conventions, "added meta test for deployment test" would become, e.g. "Add meta test for deployment test" or perhaps "Add dummy meta tag to test Netlify deployment".

PR #488, opened 2024-06-04
Replace remaining Keybase mentions with Matrix

cbeams:
Most Keybase references were removed or replaced in earlier commits, but
one outlier remained on the homepage. This commit replaces it with its
Matrix equivalent.

Note that the i18n files under _data still have many references to
keybase, but these translation strings are not in use now, and were
actually never in use on the production website. As such, I've left them
alone here, instead of expending the effort to change them when they are
effectively dead code, and I've forgone removing them, so as to save
that change for a more comprehensive cleanup that may occur later. This
commit is just a quick surgical fix.

Thanks to @kciouv for reporting the issue!


kciouv:
ACK

PR #490, opened 2024-08-27
minor text adjustment

counterweightoperator:
I find the `carry on!` statement in the current text of the Bisq2 webpage slightly confusing.

When reading for the first time, it feels as an indication to keep using Bisq 1 and not Bisq 2. Only with the rest of the sentence things become clear.

This PR changes the wording slightly to be more clear on the fact that current Bisq 1 users can install Bisq 2 without issues.


cbeams:
LGTM, thanks @counterweightoperator.

PR #544, opened 2026-05-15
Set v1.10.0

HenrikJannsen:


wodoro:
utACK

js/home.js
```diff
@@ -61,11 +68,59 @@ $( document ).ready( function() {
         return;
     }
 
+    function detectMacOSArchitecture() {
+        if( !navigator.userAgentData || !navigator.userAgentData.getHighEntropyValues ) {
+            return Promise.resolve( "mac" );
+        }
+
+        return navigator.userAgentData.getHighEntropyValues( [ "architecture" ] )
```
wodoro, line 76: getHighEntropyValues not available on safari and firefox, making it unreliable. but its okay since the fallback is showing both options at download page.

KimStrand:
utACK

